PR #1414, opened 2014-02-21
Template call

joshuaadickerson:
This makes templates infinitely pluggable by adding in 3 events for every template: pre, execute, post. You can effectively add before/after or not even call any/every template which makes themes very extensible.

It changes all calls of every template to use template_call()
If a template doesn't use template_call() it doesn't break anything so there are no backwards compatibility issues.

Although it can be added at any time, I'd still recommend some testing and code review to make sure I didn't miss anything or break anything.

See http://www.elkarte.net/community/index.php?topic=1077 for the discussion

Resubmitted to development branch from: https://github.com/elkarte/Elkarte/pull/1403

### DIFF
--- a/sources/Load.php
+++ b/sources/Load.php
@@ -1793,7 +1793,7 @@ function loadTemplate($template_name, $style_sheets = array(), $fatal = true)
 
 		// If they have specified an initialization function for this template, go ahead and call it now.
 		if (function_exists('template_' . $template_name . '_init'))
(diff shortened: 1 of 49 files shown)


joshuaadickerson:
I just thought about something. Using the event system is wrong here. The listener needs to know what the theme is so you don't call it for every theme. Then again, as an admin or developer, you might want the event to be triggered on every theme. That seems like a lot of administrative work.

PR #1762, opened 2014-08-11
Upgrade utf8

### LINKED ISSUE #1756: Add UTF8 converter to upgrade
Subject says it all.
Tracking because I already forgot once (or twice), so it's now a blocker.

emanuele45:
Here it is: UTF8 conversion while upgrading fixes #1756


emanuele45:
Since for the moment this is not necessary I'm closing it.

PR #1937, opened 2015-01-08
Some not-so-small stuff

emanuele45:
* Separated SSI into a "bootstrap" and the definition of all the SSI functions,
* moved Pbe_Imap class to sources/subs,
* fixed stuff here and there.


emanuele45:
Small bump to inform I added something I wanted to add since a long time:
http://custom.simplemachines.org/mods/index.php?mod=3104
I think it is useful in many cases.

Spuds:
Failed on 5.3 but a restart on that job cleared it up ...

emanuele45:
It's not the first time that 5.3 fails... there may be something odd going on.

PR #1950, opened 2015-01-28
Generalization of mentions + rework of scheduled tasks + introduction of modules/integration + events

emanuele45:
In other words a disaster. xD

Okay, seriously, this is a **big** pull request and I apologize for it being so big, but once I started messing with one thing, the others came in just because it looked the Right Thing<sup>TM</sup> to do and so here we are.

First, the PR is not yet completely finished, but since there is a lot going on, I'd like everyone to have a look at both the code and the general ideas behind.
The following is a quick list of stuff that has to be done before merging, there may be more that I forgot, but at least this is the "mandatory" part.
  - [x] Mentions are (still) a bit broken in regards to checks if the member can access a certain mention or not, I'd like to fix it before merging.
  - [x] Update the install and upgrade sql files to insert into the db the necessary "standard" modules (like attachments, mentions, drafts, etc.)
  - [x] <del>Remove the extends Action_Controller from Manage_*_Module classes (and fix the file name: these classes are not controllers)</del> Actuall they are controllers...well, most of them anyway.
  - [x] Fix Manage_Calendar_Module
  - [x] <del>There is no "unique" interface to enable/disable mentions (optional)</del> Being optional I took the shortest way and ignored :angel: 

Let's start with the easy ones: reorganization of scheduled tasks.
Each scheduled task is now in its own file and class and is loaded when needed, add a new scheduled task is as easy as add a file to the directory (and an entry to the database, the rest should be Just Work<sup>TM</sup> <= I'm abusing it today LOL).

Now the less easy: mentions. At the moment mentions were kind of only "mention a member in a message", "like a message", "make someone a buddy". I tried to make the system more flexible to let add more kinds of mentions/alerts (I think at some point we will have to change the name to alerts or notifications...).
Similarly to scheduled tasks I tried to split mentions to their own files, and give them a kind of standardization. Unfortunately I did that a while ago and I don't remember everything I did... lol
I'm sure I'm not done yet on that point and what I did so far is just a part of what should be done, but it's a start.

Finally modules and integration: this has become the central part of this PR. The idea behind is to "extract" code that is not strictly related to a single action and put it in another class/file that is used "when necessary". Easy.
Well, sort of.
The main convention I had to introduce is that each "module" class follows the naming scheme: ```{Name}_{Action}_Module``` where **Name** is the name of the module and **Action** is the action the module is attached to.
Management classes are instead named ```Manage_{Name}_Module```.
Following that convention, in the admin panel (in the *Core Features* page) all the modules that have a "Manage_Name_Module" class are loaded and can be enabled/disabled.
When a module is enabled it "adds" modules to an action and if necessary enables hooks (see for example the [drafts](https://github.com/emanuele45/Dialogo/blob/95f5fdf08d3a159ed478c1c15bc8c3bfe2644573/sources/admin/ManageDraftsModule.controller.php#L50) manage module class), when is disabled it removes these modules/hooks.

On the "user side", only modules enabled are loaded following the naming convention (that avoids to load a bunch of useless files).
When a module is loaded a static function [hooks](https://github.com/emanuele45/Dialogo/blob/95f5fdf08d3a159ed478c1c15bc8c3bfe2644573/sources/subs/DraftsDisplayModule.class.php#L28) is called, and this function takes care of set up the module (enabled settings and what not) and to return the configuration to set up the event listeners... yes, events are the other new thing in this PR... darn. lol

Events are a slightly more powerful way to extend ElkArte's code than hooks.
Events (at the moment) are bound to a controller, and the differences with hooks are mainly the following:
  1. the listeners are instantiated once and then shared among the events of the controller for the entire life of the controller,
  2. each event triggered can pass certain "hard-coded" variables, but it can also pass properties of the controller (either public or protected/private) and globals (just to avoid new declarations),
  3. requested variables can be passed in the order preferred by the "listener".

Something similar is applied to hooks as well (actually I'm not entirely sure I already did it.. lol), the idea is that to install an addon would be enough to extract the files into a directory and everything should work from there toggling a button in the Core Features page (or a similar one).

That's more or less a general overview of what is in this PR.
Feel free to ask any question.
his is a rather big change, and may affect the project for quite a while, so don't be afraid to hurt my feelings and if something doesn't work for you let me know, I prefer to re-do something now than have to revert huge chunks later.
Naming is something to improve (I got confused at some point and started call things with the wrong name, but now I've looked at the code for too long and I can't tell exactly what's wrong... sorry).

Well, for now that's all.


Spuds:
Are we ready to merge this one in?

emanuele45:
Unfortunately nope...
There are still at least a couple of problems: one with mentions access probably not respected (I still have to verify) and one with upgrade and install that should take care of adding some new settings.

emanuele45:
Small addition: to properly have this working I "need" the code from patch 1.0.3, so when the patch is out I'll rebase and then it will be ready for the merging... I hope. ;D

emanuele45:
I guess this is fine once the builds are finished.
There is still plenty to do, but the basics should be all in.

emanuele45:
Can I haz all the fun? :stuck_out_tongue_closed_eyes:

Spuds:
:dancers:

PR #1955, opened 2015-02-03
html5shim should use a scheme-relative URL

nwsw:
When deployed to a SSL server, the stock Elkarte theme causes security warnings due to its insecure link to the html5shim. It should use a scheme-relative URL.

My personal preference would be to avoid linking to a googlecode resource, but this change at least addresses the security issue.


emanuele45:
Yep, makes sense, thanks!

memo: to backport to 1.0.3 as well.

PR #1961, opened 2015-02-18
Minor code cleanup, added fastcgi exit for scheduled tasks.

Vekseid:
Also removed a duplicate CACHE_STALE define, and moved start time before the defines (they're slow).

### DIFF
--- a/index.php
+++ b/index.php
@@ -19,6 +19,8 @@
  *
  */
 
+$time_start = microtime(true);
+
 $forum_version = 'ElkArte 1.0.2';
 define('FORUM_VERSION', $forum_version);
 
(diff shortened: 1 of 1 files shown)


Spuds:
Thank you :+1: 

Note I'm not seeing the duplicate CACHE_STALE define other than in SSI which I don't think is avoidable, well it probably is if we move defines somewhere common.  Drat, that was a 1.02 check on my part, need to check on 1.1 as well.

emanuele45:
https://github.com/Vekseid/Elkarte/blob/d6b4aa9345178e8cf219218c569b86dcf907324e/index.php#L29
https://github.com/Vekseid/Elkarte/blob/d6b4aa9345178e8cf219218c569b86dcf907324e/index.php#L48

I probably did some mistake while moving stuff around... :angel:

Vekseid:
> other than in SSI which I don't think is avoidable, well it probably is if we move defines somewhere common.

It looks like this is what bootstrap.php should be for, for SSI, index, and install, but isn't >_>

Spuds:
Yup, my mistake, checked the wrong branch :tomato:

emanuele45:
> It looks like this is what bootstrap.php should be for, for SSI, index, and install, but isn't >_>

Yep, bootstrap.php is indeed not finished at all, I just extracted the basics and added the file in order to have "something" to work on and use it on the few files that are run as cron jobs or alike.
Use it in index.php should be the way to go.

PR #1966, opened 2015-02-21
Bringing in dev some stuff from the on-going patch 1.0.3

emanuele45:


emanuele45:
These updates are becoming a pain. lol
I think we have to update more frequently, otherwise there are good chances something is lost around.
Anyway, if you can, please, have a look if I lost something around. ;)

Spuds:
*nods* more frequent to try to make the PR to both branchs

PR #1975, opened 2015-03-05
Fix parse bbc disabled tag issue

Spuds:
add ;api for consistency with ;xml


emanuele45:
Ohh... now that I think about it: it could also work saving the value of the static variables before running the recursion and restore it later.

Spuds:
*nods* ... probably worth a quick benchmark to see which ends up being most efficient.

PR #2004, opened 2015-03-25
Auto bounce detection - Preliminary

Targren:
Post-By-Email address detects DSN (bounces), records in failed PBE table with a new, bounce-specific failure case. (Previous req Closed to fix a stupid bug)

### DIFF
--- a/sources/controllers/Emailpost.controller.php
+++ b/sources/controllers/Emailpost.controller.php
@@ -75,6 +75,13 @@ class Emailpost_Controller extends Action_Controller
 		$email_message->load_address();
 		$email_message->load_key($key);
 
+		//Check if it's a DSN
+		//Hopefully, this will eventually DO something but for now
+		//we'll just add it with a more specific error reason
+		if ($email_message->_is_dsn){
+			return pbe_emailError('error_bounced',$email_message);
+		}
+
 		// If the feature is on but the post/pm function is not enabled, just log the message.
 		if (empty($modSettings['pbe_post_enabled']) && empty($modSettings['pbe_pm_enabled']))
 			return pbe_emailError('error_email_notenabled', $email_message);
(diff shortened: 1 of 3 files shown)


Vekseid:
To clarify my last post, the following happens a lot more often than you may think

1. User requests password reset.
2. Reset mail bounces, because the account's been disabled/deleted.
3. User logs in and reactivates their account.
4. User (usually) realizes this, and requests a password reset mail again.

There may be value in setting a separate flag, and blocking the mail if too many reset requests are made in a short time, but there may be a better way to handle that.

emanuele45:
Cool stuff! (I've been a bit busy before.)

>  do you want a "dirty" flag to prevent just notification emails from going out, or "control" emails (e.g. password resets, etc..) as well? I'd personally lean towards the former, just to prevent a full lockout that needs admin intervention, but, on the other hand, the situation would really be barely different from forgetting the password and having lost access to the email account situation we've already got.

Good question.
I guess there is no reason to send out emails that we already know are not going to reach the destination, on the other hand, it could be the bounce is caused by a temporary problem, so I guess "control" emails could be send anyway.

There are a couple of tweaks to do on the coding style, but it's not really a problem, we can uniform it when integrated. ;D

Let me know if you want to refine the code before merging or if it is better to have it merged.

Targren:
This part of the code is pretty straightforward and nailed down, so I think it'd be better to have it merged, at least so we can get some more eyes on it and some more mail servers shooting at it - still have to be ready for all those exceptions. 

Sorry about the coding style faux pas. I'm trying to keep it in mind, but sometimes switching between the work style and ElkArte's isn't as seamless as I'd like. :)

emanuele45:
No problem, fix style is easy enough. ;)
I could even do it without introducing new bugs in the meantime. :laughing:

PR #2018, opened 2015-03-31
Envelope fix

Targren:
webserver user@host for envelope sender instead of the address
defined in the maillist settings/webmaster adderss. Envelope should
now match the From: address, so DSNs and replies from e.g. gmail
are properly addressed.

The issue did not appear to exist when configured to connect to an
SMTP server.

### DIFF
--- a/sources/subs/Mail.subs.php
+++ b/sources/subs/Mail.subs.php
@@ -290,7 +290,6 @@ function sendmail($to, $subject, $message, $from = null, $message_id = null, $se
 		// SMTP protocol it is
 		$mail_result = $mail_result && smtp_mail($to_array, $subject, $message, $headers, $priority, $message_id);
 	
-
 	// Clear out the stat cache.
 	trackStats();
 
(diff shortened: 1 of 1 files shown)


sources/subs/Mail.subs.php
```diff
@@ -141,7 +141,8 @@ function sendmail($to, $subject, $message, $from = null, $message_id = null, $se
 	}
 
 	// Return path, date, mailer
-	$headers .= 'Return-Path: ' . (!empty($modSettings['maillist_sitename_address']) ? $modSettings['maillist_sitename_address'] : (empty($modSettings['maillist_mail_from']) ? $webmaster_email : $modSettings['maillist_mail_from'])) . $line_break;
+	$return_path = (!empty($modSettings['maillist_sitename_address']) ? $modSettings['maillist_sitename_address'] : (empty($modSettings['maillist_mail_from']) ? $webmaster_email : $modSettings['maillist_mail_from'])); //We'll need this later
```
emanuele45: Just to nitpicking: can we haz the comment on the line before the statement and not at the end of it? :angel: :eyes: 

I know I'm a pain in the arse sometimes... sorry. :blush:

Targren: I thought I'd fixed that. Pushed the fix now.

PR #2026, opened 2015-04-08
Auto bounce detection

Targren:
Implemented new "Never" option for notifications (Magic Number 99)

Implemented code to automatically change notification_regularity to 99
on receipt of a DSN, with options to control the feature.

Outstanding needs: 
Non-standard DSNs/bounces detection is still a stub
Member is not yet notified when emails are enabled


Targren:
I still hate git. <_<

PR #2030, opened 2015-04-09
Bunch of "non functional" changes

Spuds:
Bit of cleanup here and there, fixed a couple of errors, so no really 100% non function changes, but then they were changes to areas that were non functional anyway ... its like a catch 22 !

dumping for now to see if travis et all gets pissed off.


emanuele45:
arg... something broke that one...
Can you have a look at it Spuds or should I try to rebase it?

Spuds:
I'll rebase it in a bit, going to toss a couple of more changes in this branch since there are not enough yet :tongue:

Spuds:
@emanuele45 rebase done :smile_cat:

PR #2033, opened 2015-04-13
Autoload Changes

Spuds:
This moves the autoload function to its own class file and removes it from subs.

The class structure is a bit better then what we had but its still so ugly, so many special cases it seems.

One thing I did as an experiment was to set / allow to set the set_include_path.  With that set (with SOURCEDIR, ADMINDIR, CONTROLLERDIR, ETC it then

- uses stream_resolve_include_path() in place of the if/elseif/elseif file_exists() statement since the stream command will search the include path for us.
- require_once() can just just use the basic file name without path info.  I did not change all of those statements (just some) until we decide that we are comfortable with this approach.


emanuele45:
Broken by the error handling PR... you are not lucky these days... lol

Spuds:
I figured that would happen, I'll fix it later this PM :smile_cat:

emanuele45:
Let's bring it in before it breaks again! :p

PR #2037, opened 2015-04-16
Move cleanRequest to be with the rest of Request.

Spuds:
The main thing here is moving the cleanRequest() functions that were in QueryString in to the Request class that we have, so now its with parseRequest().   

There are also a few changes just to separate out some methods, use class values,  and switch some die statements over to error::fatal_error() calls instead.

Any idea why that one is not named Request.class.php and is just Request.php

### DIFF
--- a/sources/Errors.class.php
+++ b/sources/Errors.class.php
@@ -155,9 +155,9 @@ class Errors {
 	 * @param string $error
 	 * @param string|boolean $log defaults to 'general', use false to skip _setup_fatal_error_context
 	 */
-	public static function fatal_error($error, $log = 'general')
+	public static function fatal_error($error = '', $log = 'general')
 	{
-		global $txt, $modSettings;
+		global $txt;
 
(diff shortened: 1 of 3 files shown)


Spuds:
Doh ... So I removed something :angel: and now ask do we need to add it back in? ... Travis is failing running the testYabbSeThreads testcase ... do we still need to support those?  WFT is a YabbSe :tongue: anyway it just two lines to add back but ...

emanuele45:
The only situation where it may be relevant is on converted boards from YaBB, where there are old-format links around.
I'm not sure of how frequent that happens or if it is relevant at all...
At some point it may be worth move all these "tricks" to addons (including for example the tests on passwords for other systems), I'm pretty sure some other is buried somewhere. LOL

Spuds:
*nods* ... thats kind of why I added it back in, although the comment said yabb I guess other boards may use a similar syntax, don't know so what the heck, left in in place.

PR #2040, opened 2015-04-18
Moving around files and namespaces

emanuele45:
I moved modules to their own directory and tweaked a bit the shiny new autoloader @Spuds introduced to give a basic support of namespaces so that I could use them for mention types.
Everything is still a bit messy, but things are coming together somehow.


Spuds:
Yeah ! .. namespaces :+1:

emanuele45:
Except it breaks the code validator because it surround the code, including the namespace declaration (that should be the first thing on the file) with an if... any idea how to fix it?
Strip namespaces and use?

Spuds:
Just add those files to the file exculde eval list in that file in the (testSyntaxErrors methode), I had to add that capability when I was moving some defines to constants.

emanuele45:
/me tried using shell_exec and seems to work :angel: :eyes:

Spuds:
cool idea !

emanuele45:
While messing with paths: what if classes with a namespace don't have to follow the ".class.php" or ".whatever.php" file naming patterns?
At the moment is rather inconsistent anyway (my fault for inventing fancy suffixes here and there lol), so I'd better see a more consistent naming pattern and use namespaces, that would also allow for some shorter names.

Spuds:
I agree we need either a similar or at least consistent naming pattern.  It almost does not matter *what* it is as long as the file name is consistent with the class name so the autoloader can be .... better

emanuele45:
I've been carried a little away... :angel:
I've changed all the "new" (cache, mentions, scheduled tasks) to use namespaces (that at the moment basically represent the directory they are in, so rather ugly, but that's just a first draft, there is lots of time to change it)

PR #2048, opened 2015-04-27
Errors II

Spuds:
Change from the static methods to standard class with instance loader.  @emanuele45 what do you think of doing this to update the previous static methods one I submitted?


emanuele45:
Looks :+1: to me! :smiley:

Issue #1670, opened 2014-06-28
Improve travis


emanuele45:
At the moment if the database is not setup at all, the test is passed because everything dies even before start the tests.
The first thing the script should do is test if the db is there and if not throw an error.

PR #2051, opened 2015-04-28
Upgrade

### LINKED ISSUE #944: Make the .sql files for install/upgrade unique
I'd love to have only one .sql file instead of one per db type. It would make maintenance much easier (i.e. only one file to change instead of 2/3/n when the db schema changes).

Of course we have our "abstraction layer", so we can just write the files in that "language", and then threat them as php files.

A problem may be architecture-specific changes, usually during upgrade. But if they become php files, we can find a way to run architecture-specific code at defined places.

### LINKED ISSUE #1974: Cut the bridges and remove legacy code from upgrade.php?
And with legacy I mean YaBB/SMF.

emanuele45:
Here it goes: no more YaBB/SMF in the upgrade process and a single file (plus optionally a custom one).

The only thing left now is change the ".sql" files from parse text files to proper php files.

As usual: not tested. :angel: 

Fixes #944 
Fixes #1974


emanuele45:
Do not merge yet this one even if it gives green light, it's due to #1670

emanuele45:
To me it's done.
It's not as "flexible" as before, but at least is kind of possible to understand what is going on and when (kinda).
In the past version, the error control was more tight then now, not sure if it is worth aiming at the same level at that point (probably in the future with exceptions, but for now it would be a nightmare).

I even tested both the install and the upgrade (just once) and they seem to work.
More testing is welcome, but I guess there will be more at some point. lol

Spuds:
Really .. only 445 files changed ... lazy !  LOL

Looks good, merging.

emanuele45:
LOL
Well there is also the 1.0 => 1.1 version bump included, so it's kind of normal. :yum:

PR #2056, opened 2015-05-04
Break down some of pacmans enormous functions

Spuds:
Sadly this is just a dent in a couple of pacman's painfully long and convoluted methods.  There should be no doubt this may break some package installs, so as we get 1.1 versions to install more testing can be done.

This PR is around the install and install2 functions, it creates another package file PackageActions.class.php which coordinates the work around the test and install feedback and actions. I was able to remove a lot of duplicated or very similar code in the controller and broke up a couple of the ginormous if/elseif/more/pain conditionals, but :cry: there are still many more in this beast.

Anyway it cleaner now but there is always lots of opportunity here.


emanuele45:
Nooo.... everybody loves enormous functions!! :crying_cat_face:

Spuds:
LOL .. well don't you worry then, we have a few beauty queens left :smiley_cat:

emanuele45:
Yup... profile is an example. lol

PR #2064, opened 2015-05-09
Some missing bits of the notifications system

emanuele45:
Mainly user and admin settings.

### DIFF
--- a/sources/admin/ManageFeatures.controller.php
+++ b/sources/admin/ManageFeatures.controller.php
@@ -62,7 +62,7 @@ class ManageFeatures_Controller extends Action_Controller
 	 * Mentions settings form
 	 * @var Settings_Form
 	 */
-	protected $_mentionSettings;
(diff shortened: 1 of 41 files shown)


sources/admin/ManageFeatures.controller.php
```diff
@@ -82,6 +82,9 @@ class ManageFeatures_Controller extends Action_Controller
 	public function pre_dispatch()
 	{
 		$this->_req = HttpReq::instance();
+
+		// We need this in few places so it's easier to have it loaded here
+		require_once(SUBSDIR . '/ManageFeatures.subs.php');
```
Spuds: :smile: I was thinking to do the same thing in most of those areas ... call in the base subs for the controller instead of having it in a dozen places !

emanuele45: Yeah, I feel it makes sense... at some point it could be the norm to have the base subs included by default.

PR #2073, opened 2015-05-19
Image Orienatation

Spuds:
per: http://www.elkarte.net/community/index.php?topic=2509.0

has no ACP settings, its simply On should you have imagik support.


emanuele45:
/me thinks a label "coolstuff" should be added to the tracker. LOL

PR #2076, opened 2015-05-21
Search select 

### LINKED ISSUE #2074: Alllow custom field select box to be searchable
Would it be possible to insert a custom field (select box) to be searchable into Elkarte?

http://www.elkarte.net/community/index.php?topic=2560

Spuds:
This should fix #2074

Seems to easy, so please take a look :innocent:

### DIFF
--- a/sources/controllers/Memberlist.controller.php
+++ b/sources/controllers/Memberlist.controller.php
@@ -355,7 +355,7 @@ class Memberlist_Controller extends Action_Controller
 
 				if ($this_letter != $last_letter && preg_match('~[a-z]~', $this_letter) === 1)
 				{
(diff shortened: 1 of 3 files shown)


emanuele45:
Seems to work!
BTW, Util::htmlspecialchars already defaults to ENT_COMPAT, 'UTF-8', the two could be omitted (/me adds a cleanup to the todo list :stuck_out_tongue: )

Spuds:
Glad it worked .. good catch on the excess parameters, I'll changed to Util:: to prevent double encoding and "potential" other issues.  I'll clean it the parameter list in this PR but its something to troll the code for in 1.1

PR #2079, opened 2015-05-24
fix #2077 no preview, scroll error box to top, on post2 error

### LINKED ISSUE #2077: Stop preview in case of errors (e.g. the "someone else posted in the meantime")
http://www.elkarte.net/community/index.php?topic=2566.msg17202#msg17202

Partially related to #2007

Spuds:
Removed the preview call, added a JS scroll to the top of the error box, really the category header, so its in yer face.

I did not clean up all those return $this->action_post(); which should not have the return on them, more for 1.1 unless they were done already.


emanuele45:
I missed that one!! :cry:

PR #2103, opened 2015-06-14
Development branch to 1.0.4

emanuele45:
@Spuds can you have a look if I broke anything in PackageServer.controller.php?
It was a bit hard the merge of that file... :-\


Spuds:
Not surprised about that :/  I looked at that file and it seems fine.  I'll have to got back and update it to your the post/get req class, but that is minor.  Anyway since it was not working at all, hard to make it worse :dancers:

emanuele45:
lol
Well, let's bring it in!

PR #2107, opened 2015-06-18
New style queries - fixes #2041

### LINKED ISSUE #2041: Database results
Something bothering me since a while.
A current query:
```
$request = $db->query('', 'SELECT somehting FROM {db_prefix}somewhere', array());
$data = array();
while ($row = $db->fetch_assoc($request));
{
    // maybe do something to $row
    $data[] = $row;
}
$db->free_result($request);
```

There is quite a bit of redundant stuff for our needs.
I was thinking about something like this:
```
$data = $db->queryAll('SELECT somehting FROM {db_prefix}somewhere', array());
```
the different name is there just to be backward compatible and hot have to rewrite thousands of queries in a single go.
Or maybe even better:
```
$results = $db->queryDb('SELECT somehting FROM {db_prefix}somewhere', array());
$data = $results->getData(function ($row) {
// do something
});
```
that last one could be interesting to use.

It's just an idea, so I didn't set any milestone.

emanuele45:
Well, the code still merges, so I guess I can just send the PR. :angel: 

I never planned to switch all the queries to this style in a PR, just "encourage" the use in future queries and maybe convert while going on.

### DIFF
--- a/sources/Load.php
+++ b/sources/Load.php
@@ -197,7 +197,7 @@ function loadUserSettings()
 		// Is the member data cached?
 		if (empty($modSettings['cache_enable']) || $modSettings['cache_enable'] < 2 || ($user_settings = $cache->get('user_settings-' . $id_member, 60)) == null)
 		{
-			$request = $db->query('', '
+			$user_settings = $db->fetchQuery('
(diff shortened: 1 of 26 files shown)


emanuele45:
Maybe I should make some of the commit message more "nice"?

PR #2123, opened 2015-07-15
$forum_version wasn't initialized

joshuaadickerson:
Also serves as a test of my local environment


emanuele45:
I should check before merge... LOL

$forum_version was replaced by the constant FORUM_VERSION in 1.1, probably that one was a spot that was mistakenly re-added by the merging on 1.0.4... Sorry, my fault... :blush:

PR #2141, opened 2015-07-22
Bugz and not only that! :D

emanuele45:


emanuele45:
The SQL error is not tested, so let's wait a moment to see if @vadim27 can confirm before merging. ;D

Spuds:
Looks like they have now been tested and OK'ed so I'm going to merge it in :boom:

PR #2156, opened 2015-07-26
Goodies

emanuele45:
Number of current notifications showed in the favicon and desktop notifications. :roller_coaster: :rocket: 

Plus some bug fixes as usual. :innocent:


sources/controllers/Mentions.controller.php
```diff
@@ -171,6 +171,12 @@ public function pre_dispatch()
 	 */
 	public function action_index()
 	{
+		$req = HttpReq::instance();
+		if (isset($req->query->sa) && $req->query->sa == 'fetch')
```
Spuds: I think (need to look at the code again :innocent: been lazy ) can be done as 
``` if ($req->getQuery('sa') === 'fetch') ```

emanuele45: Ohhh... that's :cool: !! :sunglasses:

emanuele45:
You didn't give me time to replace the getQuery! :P 
Ohh well, will fix later. :smile_cat:

Spuds:
Well it was fine as it was :smile: it was just a reminder about getQuery and its other getBuddies  (which can also set default values and do data sanitation)

PR #2162, opened 2015-07-29
Update index_light.css

meetdilip:
Enhancement for poll bars

New

green

Old

elk poll poll


Spuds:
Nice, much less "flat" looking :+1:

PR #2169, opened 2015-08-04
Fixes #2165 - Made dispatcher namespace-aware …

### LINKED ISSUE #2165: Impossible to call addon action controller via integrate_actions
Due to the format of the SiteDispatcher.class.php's actions array, it's not possible to call an action controller outside of the sources directory in Elkarte's core. The format needs to be changed to simply be a call to a class and method which the autoloader should take care of. This would make it possible to keep addons in their own directory under the "addons" directory.

This is the current format of the actions array.

https://github.com/elkarte/Elkarte/blob/development/sources/SiteDispatcher.class.php#L132
```PHP
// Format:
		// $_GET['action'] => array($file, $class, $method)
		$actionArray = array(
			'activate' => array('Register.controller.php', 'Register_Controller', 'action_activate'),
```

I propose this format instead:

```PHP
// Format:
		// $_GET['action'] => array('\\Full\\Path\\Class', 'action_method')
		$actionArray = array(
			'activate' => array('Register_Controller', 'action_activate'),
```

This would allow for additions to the array in the form of this for addons:

```PHP
			'page' => array('\\Elkarte\\Addon\\Pages\\Pages_Controller', 'action_index'),
```

Ant59:
… and removed dependency on file names in favour of auto-loading

I've altered the SiteDispatcher class to remove file name reliance and rely on the auto-loader instead for calling actions.

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -76,7 +76,7 @@ DEFINE('SOURCEDIR', $sourcedir);
 DEFINE('ADMINDIR', $sourcedir . '/admin');
 DEFINE('CONTROLLERDIR', $sourcedir . '/controllers');
 DEFINE('SUBSDIR', $sourcedir . '/subs');
-DEFINE('ADDONSDIR', $sourcedir . '/addons');
+DEFINE('ADDONSDIR', $boarddir . '/addons');
 unset($boarddir, $cachedir, $sourcedir, $languagedir, $extdir);
 
 // Files we cannot live without.
--- a/index.php
+++ b/index.php
@@ -80,7 +80,7 @@ DEFINE('SOURCEDIR', $sourcedir);
 DEFINE('ADMINDIR', $sourcedir . '/admin');
 DEFINE('CONTROLLERDIR', $sourcedir . '/controllers');
 DEFINE('SUBSDIR', $sourcedir . '/subs');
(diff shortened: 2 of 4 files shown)


Ant59:
Tested and working in my Vagrant environment along with my Pages add-on.

Spuds:
Thanks @Ant59 for the PR 

Looks like we just need to update the phpunit DispatcherTest.php file in the /test/sources file.  I'll try to look at that in the next couple of days, should be "easy"

Ant59:
Fixed the unit tests, and a good job too, I'd left one of the sub action 'if's to point at the now non-existent array index '2'. Also added PHPUnit to my Vagrant environment and pushed that.

I don't know if the changes I made to the test have reduced it's rigorousness, I'm still familiarising myself with unit testing. I just removed references to file names :)

emanuele45:
@Ant59 would you mind if for compatibility sake (i.e. not to rewrite addons existing) I'd change the array (and relative code) from:
```
'activate' => array('Register_Controller', 'action_activate'),
```

to:
```
'activate' => array(1 => 'Register_Controller', 'action_activate'),
```
that way, any existing addon adding something to $actionArry should still be able to (hopefully) load properly.

Ant59:
Yeah, that makes sense :)

PR #2170, opened 2015-08-05
Emails to admin and members approval

emanuele45:


emanuele45:
Another pair of eyes on the first commit (and the second) are welcome. ;D

PR #2174, opened 2015-08-09
Moar bugs

emanuele45:
@Spuds beat me in merging the pull request, so I open another one. :stuck_out_tongue:

### DIFF
--- a/SSI.php
+++ b/SSI.php
@@ -14,7 +14,7 @@
  * copyright:	2011 Simple Machines (http://www.simplemachines.org)
  * license:		BSD, See included LICENSE.TXT for terms and conditions.
  *
- * @version 1.0.4
+ * @version 1.0.5
  *
  */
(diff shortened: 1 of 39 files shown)


emanuele45:
Ouch!
@Spuds I forgot to tell you to have a look at e5ccda7a70d91415efc963222effd0f79aec7745
It's just a very simple one to suppress the error, if you prefer remove the log it can be reverted and properly fixed. ;D

Spuds:
Its fine :smile_cat: I did almost the exact same thing in one of my branches to "fix" *cough* the error.  I may have placed my version of multi implode in subs for easier reuse just in case, but other than that it was pretty much the same idea.  If there were any other tweaks I had I'll just add them in

emanuele45:
Good point on the placement... didn't thought about it... :blush:

PR #2181, opened 2015-08-12
For Review

Spuds:
Some potentially controversial :bomb: things here so proceed with caution :skull: 

1) https://github.com/elkarte/Elkarte/commit/b0f48e8928d7acea7a352ef6defc3db8b4a9e104 Fix and depreciate the permute function.  It was not returning the correct N! results, now it does but it can be dangerous to addon authors and sites in general.

2) https://github.com/elkarte/Elkarte/commit/2363cab836b490645ebec41c13d0b140b5b71dfa Add a  new permute function to allow returning of sets in place of the old function which had to compute a complete solution before returning which for sets > 3 would start to consume memory very quickly.  

This new function should be speed transparent for 99+% of real world usage,  but it will allow a larger parameter list without memory exhaustion unlike the original.  It has clamp value to prevent max execution time issues as well.  Set up to work fully through 7 parameters, the old function did not work past 3 (pre fix) and would likely run in to memory issues at 7 parameters on low cost shared hosting.

 3) https://github.com/elkarte/Elkarte/commit/e7448637adc6247c5e70e10dcdb7990965e2bd61 The "second" table repair function was over writing the keys (my dumb mistake, thank you very much) plus the allowed tag nesting definition did not seem correct.  Generally this error did not show up as preparse does most of the table repair *cough* however preparse would / could mess up tables with code in cells, and other edge cases, which this function either fixes or infuriates the user.

4) https://github.com/elkarte/Elkarte/commit/23cf58bd0261d199f015cd93d1b89b470e1c913c One thing we often see is ```[b][i][center] bla [/center][/i][/b]``` which does not render correctly since center is a block level tag :triumph:, yeah try explaining that to the end user :tomato: 

So I changed that to not be block level and used a span with display:block, yes I know, but if someone has other ideas let me know.  The change allows the above to render, as well as things like ```[size=14pt][center]TEST[/center][/size]``` etc.  

Defined as not a block for the parser simply allows it to be nested inside other tags.  The HTML that is output is span span /span /span (which I believe is valid, ie nested spans, vs a span div /div /span which would not be valid.  

5) Multi implode was on recursion passing the value as the glue, so I fixed that and added a trim function as well.

6) https://github.com/elkarte/Elkarte/commit/ffa8263452174a3bf85796fd7aa48368a698f2f6 allowed for the double encoding of entities such that they will display as the entered text.  ie ```&quot;``` will display instead of " ... Tired to address PM's and Posts only

7) https://github.com/Spuds/Elkarte/commit/017c815b10e783278051f09043f76ab03e6b62fc this is to improve the font capture and rejection when cut and paste is used in wizzy. http://www.elkarte.net/community/index.php?topic=2754

### DIFF
--- a/sources/Subs.php
+++ b/sources/Subs.php
@@ -947,9 +947,9 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 			),
 			array(
 				'tag' => 'center',
-				'before' => '<div class="centertext">',
-				'after' => '</div>',
-				'block_level' => true,
+				'before' => '<span style="display:block" class="centertext">',
+				'after' => '</span>',
+				'block_level' => false,
 			),
 			array(
 				'tag' => 'code',
--- a/sources/Subs.php
+++ b/sources/Subs.php
@@ -683,47 +683,86 @@ function un_htmlspecialchars($string)
  * - A combinations without repetition N! function so 3! = 6 and 10! = 3,628,800 combinations
  * - Used by parse_bbc to allow bbc tag parameters to be in any order and still be
  * parsed properly
- * - Caution: If passed > 8 (default) elements, will only permute the first 8 and append on the remaining elements
- * as such it will not return a proper N!
  *
+ * @deprecated since 1.0.5
  * @param mixed[] $array index array of values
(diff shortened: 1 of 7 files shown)


emanuele45:
Let's have this in!

PR #2182, opened 2015-08-14
Like counts were not updated correctly

Spuds:
When removing a post, the likes given / received counts were not updated.  

Please look at the proposed fix in like subs.  I did it more as a maintenance function, as its recounts the values, vs an update like_xxxx '-' style.   

There are a couple of complications so I'm really not sure this is the best approach, but you need to reduce the likes given for all members that liked a post that is being deleted.  You then need to reduce the likes received of the post originator by the number of likes the message had received.  To do that given an array of messages id's required a few query's, of course they only happen on delete. 

So message id -> who liked it and who posted it.  Count likes given for them and reduce by one and save.  For the poster, get the number of likes a post received and remove that from thier re-counted likes revived value.  Save that.

Also fixed the showing of liked posts that were in the stupid recycle bin

#fix2184
#fix2185

### DIFF
--- a/sources/subs/Likes.subs.php
+++ b/sources/subs/Likes.subs.php
@@ -500,3 +500,110 @@ function messageLikeCount($message)
 
 	return (int) $total;
 }
+
+/**
+ * Utility function to decrease member like counts when a message is removed
+ *
+ * When a message is removed, we need update the like counts for those who liked the message
+ * as well as those who posted the message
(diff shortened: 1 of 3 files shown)


Spuds:
Bugzzzz :bug: !

Spuds:
Fixed the postgre issue, there is still a logic issue that I need to fix so don't merge just yet

PR #2198, opened 2015-08-29
Make it possible to develop without removing the install directory

joshuaadickerson:
* add a check for $ignore_install_dir so we can develop without removing the install directory.
* Also adds a global variable, $settings_loc which is the full file path to the Settings.php file.

See: http://www.elkarte.net/community/index.php?topic=2885


joshuaadickerson:
Hmm... I should have branched. Woops.

emanuele45:
hmm...
conflicts...

Are all the commits relevant?

joshuaadickerson:
No, only the first. I didn't branch so they got added. The rest are undefined indices I found

emanuele45:
hmm... I'll have a look this evening.

emanuele45:
I replaced this with #2224, so I'm closing this one.

PR #2209, opened 2015-09-10
Front page

emanuele45:
Ohhh.... whatever!
The code to allow use any controller (that supports the feature) as front page and as a plus the starting of the release automation.


Spuds:
Is this all set to merge?  (was not sure based on your comments on the site)

emanuele45:
There is the settings bug, yes. I have to take care of it.
About functionality, I plan to add more controllers while we go, also because I don't think **many** will be suitable to appear on the front page (I think display is the only still worth probably).

emanuele45:
Rebased and fixed the bug.

Spuds:
Scruit or travis is timing on on the clover reports .. oh well

PR #2235, opened 2015-09-20
Convert PM reporting to a moderation area

emanuele45:
I'm pretty sure it's broken somewhere, but it seems to work as expected (mostly) and anyway betas are for testing, so I'm pushing it. :dancer:


Spuds:
Conflicts ! ... probably one of the other branches I merged in.

emanuele45:
I hope I fixed them! :D

PR #2248, opened 2015-09-24
Buh Bye for the most part to style="display:none"

Spuds:
Moved style="display:none" to class=hide plus associated JS to account for using css class to hide. 

a.k.a this is how you break :hankey:


emanuele45:
Cool!

PR #2250, opened 2015-09-28
Various fixes plus some more use of http req

Spuds:
A few fixes as I'm finding things we broke.

Testing out how http req is to use in the controllers.  Currently its only used in the admin area so this begins to extend its use some more.  There is still some super global use that I'm not sure about how to prevent / or if its even compatible, namely
```$this->_events->trigger('pre_load', array('_REQUEST' => &$_REQUEST, 'topic' => $topic, 'board' => &$board));``` I'm not sure why the super global is passed by reference or what manipulation is going on to them so I'm more than a bit concerned / confused over that.

Anyway want to see what the test bed says so far, probably should cherry pick out the fixes from the encapsulation as that pass by reference is not going to work as expected since the super globals are not being manipulated at that point.  So don't merge I'm afraid.


emanuele45:
Nice catch the _REQUEST.
The reason I passed it was to avoid the direct use of a superglobal in the modules (i.e. I pass a variable, it can be a superglobal, or another variable, but inside the module who cares, it's just a variable).
Now, I think it would make sense to pass the request object to the constructor of the event manager.

emanuele45:
Can be merged or you want to add moar stuff?

Spuds:
I've got a couple of more to plug away on ... just about 1/2 way through the controllers ... Currently in great pain on the PM controller :tongue:  Let me get that one committed before things are merged.  Don't want this one to conflict as its a PITA to begin with (and of course will break a thing or two).  I'l get that done this evening and then I'll be out of hand for a couple of days so you can go wild :smile:

Spuds:
Finally added the last couple I had in progress, so if your comfortable with this merge it in.  I'll do the rest of the controllers over the next week or so.  Of course this will break things as the use of $_REQUEST is so random .. sometimes it only get, sometimes only post, other times it starts as one and jumps to another, etc ... this cleans it up a bunch but I'm sure I have some wrong (plus some subs use the super globals which is another entire issue as well)

I also did some minor refact on that pm controller, mostly just the pm search area, but that entire file / controller is so fugly it just hurts to look at.   I think its close to time to break it down in to some smaller files etc, but it still just needs more breaking down of those massive methods.  Well enough whining for now !

emanuele45:
Then let's have it in!

PR #2260, opened 2015-10-08
Misc fix Part II

Spuds:
Continuing on the updating to use http req class

Several bugs found along the way and hopefully fixed

Updated some some docblocks, mostly common param details for start, per page and sort, complements of S&R so look out :innocent:   Also some spell check things etc etc.

### DIFF
--- a/SSI.php
+++ b/SSI.php
@@ -93,7 +93,8 @@ function ssi_shutdown()
  * Display a welcome message, like:
  * "Hey, User, you have 0 messages, 0 are new."
  *
- * @param string $output_method
+ * @param string $output_method The output method. If 'echo', will display
+ * everything. Otherwise returns an array of user info.
  */
 function ssi_welcome($output_method = 'echo')
 {
@@ -114,7 +115,8 @@ function ssi_welcome($output_method = 'echo')
 /**
  * Display a menu bar, like is displayed at the top of the forum.
  *
- * @param string $output_method
+ * @param string $output_method The output method. If 'echo', will display
+ * everything. Otherwise returns an array of menu data.
  */
 function ssi_menubar($output_method = 'echo')
 {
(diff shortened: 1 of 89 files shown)


Spuds:
@emanuele45 Your thoughts on this one, https://github.com/Spuds/Elkarte/commit/279db940f5fd216498f79bf17e758b15c403bcb9

This is very basic, but it allows JS to send json strings to the controllers.  This is done by posting a named variable (jsonString) the value of which is a JSON.stringify(obj) value from the script.  ( https://github.com/Spuds/Elkarte/commit/279db940f5fd216498f79bf17e758b15c403bcb9#diff-d271dde5b6ce40fd9052c434a0b4ceddL625 ) HttpReq class looks to see if its passed that sting, and then does a json_decode and places the results in the ->post object (thats not $_POST), its only in the object created in that class.  ( https://github.com/Spuds/Elkarte/commit/279db940f5fd216498f79bf17e758b15c403bcb9#diff-ee51dc566933a5ce3ce96f3c86b21ce3R133 )

Then in the controllers you can look for the post values instead of the get values like ( https://github.com/Spuds/Elkarte/commit/279db940f5fd216498f79bf17e758b15c403bcb9#diff-bf437c4bcc71653197fea8b5604e29aeR76 )

This one also updates the Suggest.class to used passed value in a constructor vs relying on context and post, nothing really to do with the above but you may see that in the mess !

emanuele45:
Looks good to me! :smile:

emanuele45:
I'm not sure if this is ready to go or you have to add more things! :innocent:

Spuds:
Feel free to merge it in ... I still have a few more controller to go through but easy enough to open part III ... plus since this still merges :innocent: thats a bonus.  Takes us up to 4.4, I'm trying to get us to 4.5 !

PR #2270, opened 2015-10-14
Hash token

Spuds:
Update most "token" areas to use a common class to generate a better random one. @emanuele45 take a look and let me know what you think.

Updates 
- session id /var
- form tokens
- form sequence
- password reset token
- 4 digit salts
- others

I also tried to consolidate using the hash function for hashes as well e.g. md5(bla) => hash('md5', bla) .. thats just for consistency.

The main purpose here is to consolidate / provide a basic token creator to replace the various sha1(md5(microtime  + mt_rand() + $modSettings['rand_seed']) + etc) crazy stuff.  

The new tokens are a-Z, 0-9 vs the old ones which were a-f, 0-9 so this adds a lot of "width" to the areas, or entropy if you wish, although it not more secure it is more crypto in its randomness vs the current way(s) they were done.

You can see this if you look at the session or tokens in the forms as they weill now have upper/lower case letters that extend to Z vs the current which limited to hexadecimal.


sources/Security.php
```diff
@@ -840,40 +822,48 @@ function createToken($action, $type = 'post')
...
  * @return boolean except for $action == 'login' where the token is returned
  */
 function validateToken($action, $type = 'post', $reset = true, $fatal = true)
 {
-	$type = $type == 'get' || $type == 'request' ? $type : 'post';
+	$type = ($type === 'get' || $type === 'request') ? $type : 'post';
 	$token_index = $type . '-' . $action;
 
-	// Logins are special: the token is used to has the password with javascript before POST it
+	// Logins are special: the token is used to have the password with javascript before POST it
```
emanuele45: Could it be the original intent of the comment was "... is used to hash the password ..."?

sources/Session.php
```diff
@@ -57,7 +57,10 @@ function loadSession()
 		// This is here to stop people from using bad junky PHPSESSIDs.
 		if (isset($_REQUEST[session_name()]) && preg_match('~^[A-Za-z0-9,-]{16,64}$~', $_REQUEST[session_name()]) == 0 && !isset($_COOKIE[session_name()]))
 		{
-			$session_id = md5(md5('elk_sess_' . time()) . mt_rand());
+			require_once(SUBSDIR . '/TokenHash.class..php');
```
emanuele45: Double dot! :P 
On a general note: isn't the autoloader able to deal with the class without having to rely on the requires?

emanuele45: Ohh... it's the file named with a double dot!

emanuele45:
Looks awesome to me!

sources/Security.php
Spuds: I'd say so :sake:

sources/Session.php
Spuds: LOL ... I'll get that fixed ... I'll try it with the autoloader, I know it missed from time to time in subs files (and I should have fixed that but did not).  I'll give it a try and see if its fatal !

Spuds:
That should fix those errors

emanuele45:
Then I guess we can just bring it in. :smiley:

PR #2277, opened 2015-10-18
Fixes again

emanuele45:
Some stuff to fix most of the current blocking issues.
@Spuds can you have a look at ebc46342082554c785b8b7372341079915c41886 and see if it actually fixes the problem? ;D

### DIFF
--- a/sources/admin/ManageSearch.controller.php
+++ b/sources/admin/ManageSearch.controller.php
@@ -132,7 +132,7 @@ class ManageSearch_Controller extends Action_Controller
 		$config_vars = $this->_searchSettings->settings();
 
 		// Perhaps the search method wants to add some settings?
-		$search = new Search();
+		$search = new \ElkArte\Search\Search();
 		$searchAPI = $search->findSearchAPI();
 
 		if (is_callable(array($searchAPI, 'searchSettings')))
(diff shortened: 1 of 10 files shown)


Spuds:
Seems to have fixed the admin search management actions, however actually performing a search results in various fatal errors.

One was in search.controller
```$this->_search = new Search();``` changing that to ```$this->_search = new \ElkArte\Search\Search();``` That results in ```ElkArte\Search\API\Errors' not found in /sources/subs/Search/API/SphinxqlSearch.class.php on line 198``` which is the static call to ```Errors::instance()->fatal_lang_error``` but its now looking inside the search namespace

**Comment by emanuele45** [comment]
hmm... my search miss the Search instantiation... dunno why...
I'll add it! :D

For the error it's "just" the matter of adding a \ in front of the class:
```
\Errors::instance()->fatal_lang_error
```
I'll check that too (and all the other classes inside them.

emanuele45:
> One was in search.controller

In addition to the already changed at https://github.com/elkarte/Elkarte/pull/2277/files#diff-4c33a887d2de4691b03e11decc3f85e4L164 ?
Where?

emanuele45:
@spuds the last commit should fix the errors reported.

Spuds:
Then lets give it a try !

PR #2290, opened 2015-10-31
Search tweaks and fixes

Spuds:
Search Stuff,  some testing of the various API's all seems to be working.  I should run results on 1.0 vs 1.1 on the same db to make sure the results have not changed but ...


emanuele45:
As long as it gives results it is fine, isn't it? :stuck_out_tongue:

PR #2297, opened 2015-11-06
Some general style and issue cleanup

Vekseid:
My IDE is still going nuts, so I figured I would address the css stuff in small chunks.

Also, setup a symlinked deployment environment. Turns out this does not work well with __DIR__

### DIFF
--- a/SSI.php
+++ b/SSI.php
@@ -303,7 +303,7 @@ function ssi_queryPosts($query_where = '', $query_where_params = array(), $query
 		censorText($row['subject']);
 		censorText($row['body']);
 
-		$preview = strip_tags(strtr($row['body'], array('<br />' => '&#10;')));
+		$preview = strip_tags(strtr($row['body'], array('<br>' => '&#10;')));
 
 		// Build the array.
 		$posts[] = array(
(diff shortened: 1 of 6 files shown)


emanuele45:
Yep, known limitation that DIR doesn't cope with symlinks, but Elk actually relies on it some way or another (see the one instance you didn't remove).
My dev environment is pretty similar to the one you use, but I have symlinked only sources and themes, index.php and the few files in the root are "real" files...

Vekseid:
Well, to fix that we could just move that logic over to Settings.php

But really I just wanted to be able to run it the way I had it.

emanuele45:
I just stripped them all from Settings.php not a long time ago based on the assumption that a settings file should just hold settings and not logic. lol

Vekseid:
Grabbing the proper install directory is a setting in my book.

Not that it matters too much, YMMV.

emanuele45:
Sorry, I've been a bit busy on other things.

> Grabbing the proper install directory is a setting in my book.

I tried to remove any login from Settings.php, because it seemed to me cleaner. In an ideal world I think the settings file could even be a yaml file for example.
Try to guess if there is around an install directory is not really a setting, more a wild guess to be sure nothing bad could happen.

But you are right YMMV. ;)

Well, I'll merge this waiting to find out what to do with ```__FILE__``` and ```__DIR__```.

Issue #2259, opened 2015-10-07
beSocial needs updating


Spuds:
Due to the template adjustments (removal of un necessary divs and classes) beSocial still needs to be updated.  This is just a reminder issue so we remember.

Spuds:
#2299

PR #2299, opened 2015-11-10
Getting beSocial up to 1.1 markup

Spuds:
Trying to update besocial to work with the damage I caused.


emanuele45:
@spuds does this one fixes #2259 ?

PR #2303, opened 2015-12-08
Josh's BBC parser

emanuele45:
I'm sending this out.
I'm pretty sure it breaks ILA and few other things, though it's probably better to have it in rather than not.

We can either work on the branch and merge it when it's done, or merge it and fix it later.
Up to you. ;)


Spuds:
I'd think it would be best to merge and then create fix PR's for what breaks, like ILA?

emanuele45:
Then I'll break everything as usual and merge! LOL

PR #2330, opened 2015-12-27
Cache

joshuaadickerson:
Working on making caching work better. Not yet implementing a caching library but at least taking some things from caching libraries to make it better. Also trying to reduce the code devoted to caching outside of the caching library

### DIFF
--- a/sources/admin/BBC.controller.php
+++ b/sources/admin/BBC.controller.php
@@ -173,8 +173,8 @@ class ManageBBC_Controller extends Action_Controller
 			Settings_Form::save_db($config_vars, $this->_req->post);
 
 			// Flush the cache so the new settings take effect
-			cache_put_data('parsing_smileys', null, 480);
-			cache_put_data('posting_smileys', null, 480);
+			Cache::instance()->put('parsing_smileys', null, 480);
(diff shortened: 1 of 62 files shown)


emanuele45:
Ouch... now this is not merging...

sources/admin/BBC.controller.php
```diff
@@ -1472,8 +1472,8 @@ public function action_setorder()
 			}
 		}
 
-		cache_put_data('parsing_smileys', null, 480);
-		cache_put_data('posting_smileys', null, 480);
+		Cache::instance()->put('parsing_smileys', null, 480);
+		Cache::instance()->put('posting_smileys', null, 480);
```
Spuds: should those be ->remove ? (and at 1691 & 1766)

Spuds:
Nice update !

sources/admin/BBC.controller.php
emanuele45: Then those in ManageLanguages.controller.php as well. :innocent:

joshuaadickerson:
Okay, I finally have all tests passing and there are no conflicts. Please merge before I break it again lol

PR #2333, opened 2015-12-29
Remove option to disable prev/next

joshuaadickerson:
http://www.elkarte.net/community/index.php?topic=2727.0

### DIFF
--- a/install/install_1-1.php
+++ b/install/install_1-1.php
@@ -2271,7 +2271,6 @@ class InstallInstructions_install_1_1
 				array('likeMinPosts', '5'),
 				array('karmaMode', '0'),
 				array('karmaTimeRestrictAdmins', '1'),
-				array('enablePreviousNext', '1'),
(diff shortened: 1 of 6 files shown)


emanuele45:
Judging from the discussion, I feel we can wait a moment and see if we can come up with a somehow better solution about the prev/next thingy.

Spuds:
should still :cherries: the bug out this this PR both here and 1.0.7

live627:
???

emanuele45:
IIRC the ```topicPointer``` issue is fixed... unless it is not.

The prev/next: the PR is here because the issue is still open: the commit here enables it by default and removes the option to turn it off. The general idea proposed on the forum was either leave it as it is or remove entirely the feature and put it into an addon.

Since there were points on both the sides, I would... keep it on hold for now and think about it for 2.0 having some better way to add it back without having to deal with code edits.

PR #2334, opened 2015-12-29
Censor

joshuaadickerson:
I think the censor should be its own class. Really, though, the reason I finally went ahead and committed this is so I could change `$user_profile` from `Array[]` to `ValuesContainer[]` and later `User[]`

You'll see a PR for this later...

From
```php
		$new_loaded_ids = array();
		while ($row = $db->fetch_assoc($request))
		{
			$new_loaded_ids[] = $row['id_member'];
			$loaded_ids[] = $row['id_member'];
			$row['options'] = array();
			$user_profile[$row['id_member']] = $row;
		}
		$db->free_result($request);
```
To
```php
		require_once(SOURCEDIR . '/ValuesContainer.php');
		$new_loaded_ids = array();
		while ($row = $db->fetch_assoc($request))
		{
			$new_loaded_ids[] = $row['id_member'];
			$loaded_ids[] = $row['id_member'];
			$row['options'] = array();
			$user_profile[$row['id_member']] = new \ElkArte\ValuesContainer($row);
		}
		$db->free_result($request);

### DIFF
--- a/SSI.php
+++ b/SSI.php
@@ -303,8 +303,8 @@ function ssi_queryPosts($query_where = '', $query_where_params = array(), $query
 		$row['body'] = $bbc_parser->parseMessage($row['body'], $row['smileys_enabled']);
 
 		// Censor it!
-		censorText($row['subject']);
-		censorText($row['body']);
+		$row['subject'] = censor($row['subject']);
+		$row['body'] = censor($row['body']);
 
 		$preview = strip_tags(strtr($row['body'], array('<br>' => '&#10;')));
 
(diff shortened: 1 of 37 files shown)


emanuele45:
Ouch I used the wrong file name! (forgot the .class)
Anyway, the "require_once" should not be necessary. :P

joshuaadickerson:
It doesn't work without the cache

joshuaadickerson:
"class"

On my phone

joshuaadickerson:
If ".class" isn't there, it doesn't work without including the file.

emanuele45:
Odd, I used it in few places, but never require'd the file... :confused: 
Ohh well, who cares.

PR #2335, opened 2015-12-31
Add original poster block to ssi_recentTopics query

kode54:
This change to the SSI script makes way for changes to Simple Portal, and possibly other scripts, to display the topic starter's name and profile link.

Eventually, I may make further changes to include such info as the timestamp of the opening post. All the other info I need is already provided, though.


Spuds:
Thanks .... its good information for that function to return.

PR #2339, opened 2016-01-02
Php7

emanuele45:
And finally added support for PHP 7 in tests (and ElkArte as well, of course :P).
The problem were two:
1) session handling was broken, thanks to @albertlast (and his pull request at SMF) for putting me on the right track,
2) the eval function has a different behaviour that requires a try/catch breaking the php syntax validation function


emanuele45:
It passes the tests so I guess we can bring it in. :stuck_out_tongue:

PR #2340, opened 2016-01-02
Formatting

emanuele45:
Just some formatting (spaces to tabs) and a new test to check (c) headers are there and are formatted more or less the way we love them.

### DIFF
--- a/addons/index.php
+++ b/addons/index.php
@@ -1,0 +1,16 @@
+<?php
+
+/**
+ * This file is here solely to protect your addons directory.
+ */
+
+// Look for Settings.php....
+if (file_exists(dirname(dirname(__FILE__)) . '/Settings.php'))
+{
+	// Found it!
+	require(dirname(dirname(__FILE__)) . '/Settings.php');
+	header('Location: ' . $boardurl);
+}
+// Can't find it... just forget it.
+else
+	exit;
--- a/install/CommonCode.php
+++ b/install/CommonCode.php
@@ -1,6 +1,21 @@
 <?php
 
 /**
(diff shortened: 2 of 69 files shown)


Spuds:
Glad to have another code :trollface: is in da house !  :tada:

emanuele45:
I was thinking about implementing an automated :trollface: for testing indentation, but is too much of an hassle.

Spuds:
:trollface: : saz ... better to spend time on things that look like this instead :tongue: 
```
}

public function WhereAreMeDocBlocks(l$ots, $of, $params)
{
}
```

joshuaadickerson:
Answer: New pedantic tests for formatting
Question: What happens when you submit a PR with spaces instead of tabs and missing headings?

I finally fixed PhpStorm so now it is following my settings. It even tells me now that files are indented with spaces instead of tabs and asks me if I want to fix them. I click a single button and it does it for me.

emanuele45:
> Question: What happens when you submit a PR with spaces instead of tabs and missing headings?

Space <=> tabs: the blame history of the file is mostly broken because
that single commit changes the entire file (and probably moving the
file as well .
Yes, it should be the responsibility of the "merger" to ensure it
doesn't break, but people can do mistakes (/me is :innocent: lol) and
miss the problem.

Missing headers: well, in theory nothing, in practice inconsistency,
that is bad by itself. :P
If we were to consider moving from "header in each file" to "(c) in
LICENCE.txt" we'd have to face the fact SM requires explicitly that
SMF headers should be in each file, so we are not allowed (due to the
precedents with other projects we already know that).

PR #2345, opened 2016-01-09
Moving color css to _light and non-color css out of it. 

Vekseid:
Some minor attempts at trying to resolve the current css mess. : )


Spuds:
:+1: Thanks!

PR #2349, opened 2016-01-14
Added support for the class-only interface of the current PECL memcac…

kode54:
…hed module, which can be built for PHP 7.


emanuele45:
Thanks! :D

Spuds:
:+1:

PR #2350, opened 2016-01-15
Only emit session id in query parameters when cookie is not set.

kode54:
This changes the regular expression replace function used by the queryless URLs option, so that it doesn't insert the SID parameter if a cookie is set.

### DIFF
--- a/sources/QueryString.php
+++ b/sources/QueryString.php
@@ -289,11 +289,10 @@ function ob_sessrewrite($buffer)
 function buffer_callback($matches)
 {
 	global $scripturl;
-	static $robotset = false;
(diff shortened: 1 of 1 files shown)


kode54:
This seems to work properly, but I still see an occasional error in the termination of administrator sessions, where it ends up leaving a stray PHPSESSID= between the index.php and the trailing board, or topic,.

emanuele45:
Sorry for the long time waiting...I lost the email... :blush:

Spuds:
For some reason I never saw the notification (that the static had been removed), or I would have merged this ... Sometimes Github is strange.

PR #2351, opened 2016-01-15
Fix avatar resizing for guests browsing the forums.

kode54:
Add .avatarresize style bit for guests browsing the forums.

### DIFF
--- a/sources/Load.php
+++ b/sources/Load.php
@@ -295,6 +295,27 @@ function loadUserSettings()
 	// If the user is a guest, initialize all the critical user settings.
 	else
 	{
+		// Guests need this too, it won't be emitted otherwise
+		static $added_once = false;
+
+		if (!$added_once)
+		{
+			if (!isset($context['html_headers']))
+				$context['html_headers'] = '';
+
+			if (!empty($modSettings['avatar_max_width']) || !empty($modSettings['avatar_max_height']))
+			{
+				$context['html_headers'] .= '
+	<style>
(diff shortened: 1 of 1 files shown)


emanuele45:
I'm a bit surprised of this (unless it's a bug in 1.1 that I can't test right now), on elkarte.net it seems to work fine and the relevant code is added to the template both for logged in members and guests (i.e. browsing the forum as a guest).

Can you give me some instructions on how to reproduce the problem?

kode54:
It needs for avatar size limits to be enforced only as CSS resizing, not re-encoding or outright disallowing.

emanuele45:
So, I have the forum (1.1) set to "let CSS resize them (avatar)":
screen_430

and I'm browsing as a guest:
screen_430
but here I see the "resize" code block for the resize (see the bluish area of the Chrome inspect on the right).

kode54:
It must only affect 1.0.6, then. Close and drop.

kode54:
It also depends on user avatars on the server being noticeably larger than 65x65, the default size.

emanuele45:
Let's keep it open to be tracked as a possible bug, it may be in a certain combination it may fail.

emanuele45:
This was a very valid report, sorry @kode54 for dismissing it too quickly.
I pushed here another fix:
https://github.com/elkarte/Elkarte/pull/2458/commits/2210a08ecd96325710895f499e2ade30f178bb1e
I moved entirely the original code up in the chain to loadTheme where it better belongs.

PR #2352, opened 2016-01-16
Added support for Nginx queryless URLs.

kode54:
This still needs a modification to the documentation, as setting this up for Nginx requires a slight bit of configuration changes:

Presumably it needs the fastcgi_split_path_info argument in the following example PHP config block:

```
    location ~ \.php$ {
        try_files $uri =404;
        fastcgi_split_path_info ^(.+\.php)(/.+)$;
        include fastcgi_params;
        fastcgi_index index.php;
        fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
        fastcgi_read_timeout 120;
        fastcgi_pass unix:/var/run/php-fpm/php-fpm.sock;
    }
```

And it also may need this:

```
    location / {
        try_files $uri $uri/ /index.php?$args;
    }
```


Spuds:
This is the php section that I generally use (note I use separate socks per "site" )
```
	# regex to split $uri to $fastcgi_script_name and $fastcgi_path
	fastcgi_split_path_info ^(.+\.php)(/.+)$;

	# Check that the PHP script exists before passing it
	try_files $fastcgi_script_name =404;

	# Bypass the fact that try_files resets $fastcgi_path_info
	# see: http://trac.nginx.org/nginx/ticket/321
	set $path_info $fastcgi_path_info;
	fastcgi_param PATH_INFO $path_info;
	
	# Change YOURSITE to the pool or to php-fpm if using a global one
	fastcgi_pass unix:/tmp/php5-YOURSITE.sock;
	fastcgi_index index.php;
	include fastcgi_params;
}
```

I've never tried the queryless urls on NGINX, I don't _think_ they work with out the "location" directive in the appropriate sites-available file (as you noted).  This is what is different from Apache which consumes those queryless ones w/o anything being done to .htaccess   I think this is why it was not included for NGINX vs Apache.

If we add this I think we should also provide a working set of directives for the user to add.  I'm not sure what NGINX would do without the right directives in place when you check that box, it may render the forum defunct/404 all requests?

So to do this we need the rewirte rules that would work, and maybe a bit fat warning about checking that box if you don't have them in place first (on NGINX)
```
	location / {
		try_files $uri $uri/ @rewrites;
	}

	location @rewrites {
		# Place any rewrite rules here
		rewrite ^ /index.php last;
	}
```

However if there are no ill side effects, then all we would need is a recommended set of rewrites for them to use.

kode54:
Lack of proper rewrite rules only breaks direct links to boards and topics, which are the only thing rewritten at this time, last I checked. Also, there seems to be a bug with rewriting the link that handles ending an administrator session, or at least the redirect that follows logging in or out of an administrator session.

1. Logging out of an administrator session may redirect to a link that is like $scripturl . '?PHPSESSID=sessoncode&amp;/board,#.html'
2. Being redirected to an administrator session password prompt from a page already containing the same password prompt, redirects right back to that password prompt after successfully inputting a password.

emanuele45:
So is this one fine?

And @kode54 could you open different reports for the bugs?

PR #2353, opened 2016-01-16
Fix all variable variables for PHP 7

kode54:
This hopefully fixes all cases of potential ambiguity with "variable variables", where a variable is used to reference a member of another structure or class. A lot of cases were probably clear, with using plain variables. Others, where the key was referenced from an array, would break on PHP 7 if not referenced:

$class->{$array['key']}

### DIFF
--- a/sources/Action.controller.php
+++ b/sources/Action.controller.php
@@ -190,7 +190,7 @@ abstract class Action_Controller
 	{
 		if (property_exists($this, $dep))
 		{
-			$dependencies[$dep] = &$this->$dep;
+			$dependencies[$dep] = &$this->{$dep};
 		}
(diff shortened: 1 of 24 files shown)


kode54:
I notice there were a lot more "variable variables" in the development branch than in release 1.0.6.

Spuds:
Ah, cool.  Did not know you had to do that with php 7, thanks !

Nao:
From what I can see, this change is only needed when you're accessing an array or object from that variable variable.
e.g., $obj->$arr['hello'] needs to be turned into $obj->{$arr['hello']}, but $obj->$hello can remain untouched.
(Just my 2 cents.)

PR #2358, opened 2016-01-24
Fix random things

Spuds:
Still trying to prep a 1.1 B1  ... needs moar work


emanuele45:
heh...
Apparently the testing of the headers is slightly too picky...
It should be enough to test for \s (spaces) instead of the exact sequence of spaces and tabs:
https://github.com/elkarte/Elkarte/blob/development/tests/sources/FilesTest.Basic.php#L110
https://github.com/elkarte/Elkarte/blob/development/tests/sources/FilesTest.Basic.php#L111
and so on... would you like to do the honors? :innocent:

Spuds:
LOL ... I was doing just that when you entered the comment ... I used /s+? which should cover spaces and tabs, any number, any combo ... lets see what it thinks now

Spuds:
Worked .. Travis gave me a :trophy:

emanuele45:
In that one as well! :tada:

PR #2359, opened 2016-01-27
Fix adding session ID to redirects if a cookie is still set.

kode54:
This fixes a bug which would occur with queryless URL generation, where a session code would be inserted between the page name and the /parameters added later, which no longer handles the SID parameter if a cookie is set.


Spuds:
I have no idea why the header check fails only for PHP7 ....

Spuds:
Travis error should be fixed with this commit: https://github.com/elkarte/Elkarte/pull/2363

PR #2362, opened 2016-01-28
Update BB to 2.2.18

Spuds:
Minor update + php7 compatibility


Spuds:
The failure is only on php7 header checks ... no idea why.
```
1) TestFiles::testHeaders

/var/www/sources/subs/SettingsForm.class.php

Failed asserting that false is true.

/var/www/tests/sources/FilesTest.Basic.php:238
```

Spuds:
Travis error should be fixed with this commit: https://github.com/elkarte/Elkarte/pull/2363

PR #2363, opened 2016-01-28
Travis Updates

Spuds:
- Fix catastrophic recursion for php7 and header check
- DROP PHP 5.4 and PHP 5.5 testing, 
- Replace above with 5.6
  - Now the matrix is 5.3, 5.6, 7.0
- Add MySQL 5.6 and 5.7 testing.  Travis only had MySQL 5.5 installed, so to do this you have to install the different versions, you can see how that was done in the setup-server.sh
  - PHP 5.3 / MySQL 5.5
  - PHP 5.6 / MySQL 5.6
  - PHP 7.0 / MySQL 5.7
- Add postgre 9.4 testing for PHP 7.0 .. Travis did have this installed, so I enabled it for 7.0.  If you want later versions they would have to be installed like I did for MySQL.  The other tests use 9.3
- Move code coverage to the 5.6 test run (was on 5.4)


kode54:
What about MariaDB? Maybe not worth testing, since it covers the same required feature set anyway.

KeiroD:
@kode54 MariaDB is taken care of by MySQL tests. MariaDB is drop-in replacement for MySQL.

PR #2367, opened 2016-02-02
A bunch of stuff for ILA and drafts

emanuele45:
It mostly works, I think.
It may be good enough for a beta even though it's broken (there is (still) a race condition for sure).


Spuds:
Lets bring it in then!

PR #2390, opened 2016-02-24
Maillist IMAP heading to info box

LouAugust:
This corrects the double heading display of the IMAP server. Final result is one heading followed by an information box. 
Relevant post at http://www.elkarte.net/community/index.php?topic=3324.0


Spuds:
:+1:

Issue #2393, opened 2016-02-26
Add defaults to custom fields


Vekseid:
This is the most obvious thing that Personal Text / Blurb can do that custom fields currently can't. Besides being shown without title, which I've already got done in my own code.

emanuele45:
I thought it was already there, but it not!... heck I know close to nothing about custom fields! xD

Anyway, just to be kind of consistent, take in consideration that in 1.0.6 I did some changes to radio and selects regarding the default value https://github.com/emanuele45/Dialogo/commit/6d010104884895e1121f29a36587353b06ebaa70 
It's not yet merged into development, I'll bring it soonish.
It shouldn't conflict with this, but just in case if there is anything that may be re-used. ;)

Vekseid:
@emanuele45 you closed this but didn't merge the pr?

emanuele45:
Darn github! xD
I forgot development bugs are closed automatically...

Issue #1322, opened 2014-01-29
Gender Selection


Xarcell:
Allow the choice between 3 genders. Male, Female, and Intersex. several countries recognize a third gender like Intersex, such as: Australia, Germany, India, Nepal, and Bangladesh.

We are talking about those that are born with both characteristic's or genitally of male/female. Not those that are transgender, who switch genders or cross-gender.

I would also like to see gender be directly tied to membergroups, so that permissions can be placed on user's based on their gender.

Of course, then you would have make it so that a user's gender can only be "changed" by the admin or global moderator.

Vekseid:
Intersex isn't a gender, it's a physiological condition.

Genders could be male/female/both/none/other/undisclosed

Since even Facebook supports this these days, I think this can be supported more directly (if only by permitting admins to define more fields).

Spuds:
I wonder if that standard field should be removed an just left for the admin to use a custom profile field instead then.  Could seed it with your suggestion and and admin can add / remove as desired.  Not sure what to do in the gender icons though.

Xarcell:
@Spuds if you make gender a custom profile field, then so should all other fields. Which might not be a bad idea when you think about it. What would be the pros and cons of it? Pros: Leaner code, less maintenance, smaller package, ... Cons: people bitching it's not there by default, people bitching that they don't understand how to create profile fields, people bitching that it was changed...

If it were me, all profile fields would have to be created via admin. However, how would this effect mods that check for gender?

Spuds:
@Xarcell good to see you around !

Honestly I'm not a big fan of the default profile fields and have long thought about removing them all as they are just kind of messy in the code.  They are nice for noobs who can find the custom fields intimidating on first use. 

I'd think we could just preload additional "default" custom fields at the start and cover whats there, or even have a small library of templates that gets json fetched that you could select from.  

Probably right about addons that use gender or any of the standard fields, the ones that remain are kind of dug in.  Perhaps the use of reserved names (ugh) for them so at least addons could find the ones that were typically around,  have not looked at that code area in a bit so that may not be necessary.

Vekseid:
@Xarcell why would there be cons? You could just have a default set of custom fields.

@Spuds custom fields should already be able to reference images, shouldn't they?

Vekseid:
@Spuds @emanuele45 I noticed you moved this to 2.0 but gender is something I kindof want to take care of. Like, this weekend. >_>

[...]

PR #2394, opened 2016-02-27
This removes the location, gender, and personal text custom fields

Vekseid:
also adds an 'above icons' custom field location to enable mimicking old, slightly more flexible behavior.

This is part one of a three-stage resolution for #1322. It just touches a lot of files so I wanted to merge one component at a time.

'Website' unfortunately still remains as a standard field, and 'custom title' is also a bit suspect. A proper custom personal text replacement would also need defaults as described in #2393.


Vekseid:
Ugh, looks like I missed something when altering the member table. Hrm

Vekseid:
That insert format is so obnoxious I am tempted to just rewrite it. I am not sure why it's failing since this is the error I solved with the latest commit, though. : /

emanuele45:
I think you missed the one in ```tests/travis-ci/setup.php```, there you removed only the fields from the first array and not from the second.

Vekseid:
Well that was embarrassing.

Is there some reason the install script (and its test) do the member insert like that?

Vekseid:
Any further comment on this?

emanuele45:
The only thing I do not see is the code to deal with the upgrade process.
Do we want to merge and deal with that part later?

Vekseid:
That's my current plan. #1322 needs icon work (separate) and if we can do something for #2393, that changes what the upgrade path would look like a bit.

emanuele45:
Ohh, okay I see it.

PR #2415, opened 2016-03-05
Some fixes to previous code, more icons, some cleanup.

Vekseid:
Note that smc_Popup now expects a string for the icon rather than a filename (referencing an appropriate i- icon).

So much code duplication. We need to have a talk. -_-


Vekseid:
Feel free to merge this one. The next task for this is a rather nuts amount of work.

Issue #2413, opened 2016-03-05
1.1-dev shows users as being logged in forever


Vekseid:
At least while using database sessions.

Vekseid:
I am not sure what is going on with this one. There doesn't seem to be anything that gets rid of old log_online entries?

emanuele45:
Do you mean in the who-is-online page?

Vekseid:
Everywhere - whether or not someone is 'online'. Their is_online property is true.

No proper pruning of the log_online table seems to be occurring and that is where the software is getting the is_online status from.

emanuele45:
The issue is in:
```
$do_delete = $cache->getVar($do_delete, 'log_online-update', 30) && (int) $do_delete < time() - 30;
```
if cache is disabled getVar returns false, so $do_delete is never true and the log_online never cleaned up.
If I'm not completely :sleeping: I think in this case test if the cache was hit is useless, jsu:
```
$do_delete = $cache->get('log_online-update', 30) < time() - 30;
```
should suffice:
* cache enabled a value is returned and compared to the last 30 seconds gap,
* cache disabled an equivalent to 0 is returned and the delete will always be applied.

PR #2419, opened 2016-03-07
Adds defaults to custom fields, closes #2393

### LINKED ISSUE #2393: Add defaults to custom fields

Vekseid:


Vekseid:
I am curious how my test forum still ran with that.

sources/Load.php
```diff
@@ -1020,7 +1020,7 @@ function loadMemberContext($user, $display_custom_fields = false)
 	// Set things up to be used before hand.
 	$profile['signature'] = str_replace(array("\n", "\r"), array('<br />', ''), $profile['signature']);
 	$profile['signature'] = $parsers->parseSignature($profile['signature'], true);
-	$profile['is_online'] = (!empty($profile['show_online']) || allowedTo('moderate_forum')) && $profile['is_online'] > 0;
+	$profile['is_online'] = !empty($profile['show_online']) || (allowedTo('moderate_forum') && $profile['is_online'] > 0);
```
emanuele45: Is this related to #2413 ?

Vekseid: Actually that's the wrong grouping. Ack.

And not exactly, I saw the useless parens while trying to figure out what was going on with it.

PR #2426, opened 2016-03-09
Genderblending - ready!

Vekseid:
Still need to finalize things for the icons and add everything to the upgrade script.


Vekseid:
Alright, good to go @Spuds @emanuele45 

#1322 was not something I could afford to have put off. >_>

install/upgrade_1-1.php
```diff
@@ -554,6 +554,101 @@ public function custom_field_updates()
...
+						    ));
+				}
+			),
+			array(
+				'debug_title' => 'Updating icon custom fields',
+				'function' => function($db, $db_table) {
+					$result = $db->query('', 'SELECT id_field, col_name FROM {db_prefix}custom_fields WHERE placement=1');
+
+					while ($row = mysqli_fetch_assoc($result))
+					{ // <a href="skype:{INPUT}?call" class="icon i-skype icon-big" title="Skype call {INPUT}"><s>Skype call {INPUT}</s></a>
```
emanuele45: Is that a reminder or a left over? :angel:

themes/default/ProfileInfo.template.php
```diff
@@ -771,8 +771,10 @@ function template_profile_block_contact()
 					$cf_show = true;
 				}
 
+				// removed $field['name'], ': ', as it was annoying
+
 				echo '
-					<li>', $field['name'], ': ', $field['output_html'], '</li>';
+					<li>', $field['output_html'], '</li>';
```
emanuele45: Did you completely remove the name of the field from the profile? (Question, just to understand if it was intended or if I missed something around :wink: )

install/upgrade_1-1.php
Vekseid: Well at least it's not a bug >_>

themes/default/ProfileInfo.template.php
Vekseid: Intended. It was horrifically ugly.

I just made sure all of the icons have accessible text and title attributes instead. Otherwise, screen readers will often read it twice anyway, and sighted people will recognize most of the icons as-is.

PR #2431, opened 2016-03-11
Some low-hanging icon fruit

Vekseid:
Stuff that was simple enough to get out of the way for beta 1. : )


emanuele45:
Maybe just dump everything in the template? (And keep the index definition there just to avoid undefined indexes around.)

Oh, and that reminds me that the sorting template could very well be split out in a dedicated template function. :angel:

Vekseid:
Yeah there are a number of functions like that. Some things I am treading on comparatively lightly at the moment because I've got ideas for how to generify them.

Vekseid:
Going to try to get a few more icons in.

emanuele45:
> Yeah there are a number of functions like that. Some things I am treading on comparatively lightly at the moment because I've got ideas for how to generify them.

Probably the best would actually be to completely unify the two templates... well, at least some.

Vekseid:
The best would be a full refactoring. That's a post-1.1 discussion though I think.

Vekseid:
So many things are... creatively embedded into the code. : /

PR #2432, opened 2016-03-12
Updates for very long board names breaking layout

Spuds:
Discussed on the site .. http://www.elkarte.net/community/index.php?topic=3358

Not sure we want to fix this for 1.0.7 but here it is anyway

Board name with no natural break points, losses alignment on board index and overflowed the select drop down.  This will partially fix the issue should it ever arise in the real world (the select pull down can still misbehave)


emanuele45:
Why not? :angel:

PR #2447, opened 2016-03-15
Fixes

emanuele45:
Some random stuff for 1.1, mostly from the quick public testing run.

@joshuaadickerson see if https://github.com/elkarte/Elkarte/commit/9dfeeea9b22a82be4a4d6a2e1e6788fddaf0f883 makes sense to you. ;)


Vekseid:
Was about to report the email template thing as a bug : p

PR #2452, opened 2016-03-19
Revert attachments in draft

emanuele45:
Sad, but I give up.
Save attachments for drafts is something that is far beyond the scope and possibilities of 1.1 unless we want to redo a lot of stuff.

In this PR I reverted the changes I made to try to save attachments for drafts as well, fixed the ILA (including preview) and added few default mime showing the appropriate icon.

Still some goodies are missing (like alignment, resizing, etc., but they can wait a bit, even a B2.

### DIFF
--- a/install/install_1-1.php
+++ b/install/install_1-1.php
@@ -65,7 +65,6 @@ class InstallInstructions_install_1_1
 				array('name' => 'id_msg',          'type' => 'int', 'size' => 10, 'unsigned' => true, 'default' => 0),
 				array('name' => 'id_member',       'type' => 'mediumint', 'size' => 8, 'unsigned' => true, 'default' => 0),
 				array('name' => 'id_folder',       'type' => 'tinyint', 'size' => 3, 'default' => 1),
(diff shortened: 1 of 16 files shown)


Spuds:
:crying_cat_face: sorry you could not get it working, I know how that feels! ... what was the final straw on the :camel:  ?

emanuele45:
Basically security.
Having to deal with attachments not associated to a topic saved in the database means they need another field "owner" or something similar to check against, so yet another mess of queries to change to test for ownership as well. A mess...
I think there is to re-think a bit the attachments database structure to have a higher level of flexibility (i.e. ditch the avatars from the attachments table, so that id_member becomes not something we rely on in order to find the avatar of a member, ditch the thumbnails, and probably more).

Oh, do not merge it yet, I missed some stuff!

emanuele45:
Okay, that commit should do.

Vekseid:
Yeah the way SMF/Elkarte/MyBB/phpBB handle attachments is so bad I just turn it off for any forums I run in the first place. >_>

PR #2453, opened 2016-03-20
Backport of logging referring URL - FIxes 2435

emanuele45:


Spuds:
Nothing to do with this commit, but should we update parse_bbc with nofollow for all external links?  Seems thats ipso facto standard.  Then addons like yours can enable it if wanted.

PR #2454, opened 2016-03-20
Some problems with like stats

emanuele45:


Spuds:
We may consider moving this to stats in b2 ... I'm just not convinced its top menu worthy.

emanuele45:
Yep, I think there was even an issue somewhere.

Vekseid:
Was wondering why it had a top level menu item >_>

Spuds:
The author of the original commit did it that way and it was never changed.  I was going to move it when I did some work on the functions but ran out of momentum.  We should get it sorted in the betas.

PR #2462, opened 2016-03-25
Youtube timestamps

Frenzie:
Added support for YouTube timestamps and refactored the code a little to get rid of redundancy.


Frenzie:
The commit I just added fixes up an oversight on my end (I hadn't added timestamps to double-click), but also I realized that `'?amprel=0'` was nothing but a severely botched form of `&amp;rel=0`. Since this can't have been working properly in that form, ever, I suppose technically that's more of a change than a fix. Specifically, `rel=0` doesn't show related videos when playback is finished, while `rel=1` (the default) does. So if you prefer it the way it's been working, I'll just remove the `rel` parameter altogether instead.

Spuds:
I feel what you did is proper ... rel=0 ... good catch on the botched &amp;, thanks for your work

PR #2477, opened 2016-03-31
Queryless

emanuele45:
More or less an experiment.
If you don't like the style, let me know and I'll drop the controversial changes. ;D

### DIFF
--- a/sources/Load.php
+++ b/sources/Load.php
@@ -2490,44 +2490,29 @@ function determineAvatar($profile)
 function detectServer()
 {
 	global $context;
+	static $server = null;
 
-	$context['server'] = array(
-		'is_iis'			=> serverIs('iis'),
-		'is_apache'			=> serverIs('apache'),
-		'is_litespeed' 		=> serverIs('litespeed'),
-		'is_lighttpd' 		=> serverIs('lighttpd'),
-		'is_nginx' 			=> serverIs('nginx'),
-		'is_cgi' 			=> serverIs('cgi'),
-		'is_windows' 		=> serverIs('windows'),
-		'iso_case_folding' 	=> serverIs('iso_case_folding'),
+	if ($server === null)
+	{
+		$server = new Server($_SERVER);
+		$servers = array('iis', 'apache', 'litespeed', 'lighttpd', 'nginx', 'cgi', 'windows');
+		$context['server'] = array();
+		foreach ($servers as $name)
+		{
+			$context['server']['is_' . $name] = $server->is($name);
+		}
+
+		$context['server']['iso_case_folding'] = $server->is('iso_case_folding');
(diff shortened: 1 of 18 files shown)


sources/Load.php
```diff
@@ -2488,44 +2488,29 @@ function determineAvatar($profile)
...
-		'is_apache'			=> serverIs('apache'),
-		'is_litespeed' 		=> serverIs('litespeed'),
-		'is_lighttpd' 		=> serverIs('lighttpd'),
-		'is_nginx' 			=> serverIs('nginx'),
-		'is_cgi' 			=> serverIs('cgi'),
-		'is_windows' 		=> serverIs('windows'),
-		'iso_case_folding' 	=> serverIs('iso_case_folding'),
+	if ($server === null)
+	{
+		$server = new Server($_SERVER);
```
Spuds: Why pass the super global?

sources/admin/ManageFeatures.controller.php
```diff
@@ -1335,9 +1335,6 @@ private function _basicSettings()
 				array('text', 'jqueryui_version', 'postinput' => $txt['jqueryui_custom_after']),
 				array('check', 'minify_css_js'),
 			'',
-				// SEO stuff
-				array('check', 'queryless_urls', 'subtext' => '<strong>' . $txt['queryless_urls_note'] . '</strong><br />' . ($context['server']['is_apache'] || $context['server']['is_nginx'] || $context['server']['is_lighttpd'] ? $txt['queryless_urls_work'] : '<span class="error">' . $txt['queryless_urls_notwork'] . '</span>')),
-			'',
```
Spuds: :+1:

Spuds:
Makes sense to me ... kind of like the browser class ... did http_req also have server loaded?  I should remember but don't !

sources/Load.php
emanuele45: Fanciness? O:-)
Actually a piece is missing (i.e. check if the variable is set or not).
```php
if (isset($_SERVER))
    $server = new Server($_SERVER);
else
    $server = new Server(array());
```

I guess I can use $_SERVER by default and in case let it accept an argument to override or nothing if $_SERVER doesn't exist?
/me likes having overrides when possible. :angel:

emanuele45:
Yep, it has, and it was for that reason I started putting this together, in particular because of:
http://www.elkarte.net/community/index.php?topic=3438.0

sources/Load.php
Spuds: I'd go for that ... just a ``` $server = new Server();``` and then pass the logic to the constructor, so same thing different location.
```
protected $__server = array()l;

public function __construct($server = null)
{
if (is_array($server))
$this->__server = $server;
else
$this->__server = $_SERVER;
}
```

emanuele45:
Dammit... it doesn't merge... :crying_cat_face:

emanuele45:
Rebased.

PR #2485, opened 2016-04-04
Drop harmful sanitizeMSCutPaste() function

Frenzie:
Alternatively you fix it with something like `$replacechars = array('&sbquo;', '&bdquo;', //etc.` but that'd be a meaningless exercise in futility.

### DIFF
--- a/sources/Subs.php
+++ b/sources/Subs.php
@@ -1410,51 +1410,6 @@ function remove_integration_function($hook, $function, $file = '')
 }
 
 /**
- * Microsoft uses their own character set Code Page 1252 (CP1252), which is a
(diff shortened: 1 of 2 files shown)


Frenzie:
See http://www.elkarte.net/community/index.php?topic=3272.0 and http://www.elkarte.net/community/index.php?topic=3451.0

The characters this function so inaptly replaces are real UTF-8 encodings. See e.g. http://unicode-table.com/en/2014/ There's nothing "MS" about them. It may be different in ISO-8859-1 ([see here](https://github.com/SimpleMachines/SMF2.1/blob/97876e73732658f41566808a7678fa0083cf2cd3/Sources/Subs.php#L4547-L4557)) but we don't care about that since Elk is exclusively in UTF-8.

PR #2493, opened 2016-04-14
Don't cause useless YT 301 redirect

Frenzie:


Spuds:
Not sure why scruitinizer is failing here .. does not seem to be anything to do with the PR

PR #2497, opened 2016-04-19
Errors

emanuele45:
First little round of pre-beta bug fixes.


Spuds:
I'll leave you to merge this since I don't know if you are done.

emanuele45:
Let's find a solution for the previews and I think it can be merged. ;D

emanuele45:
Spuds wrote on 19/04/16 22:47:
>
> I'll leave you to merge this since I don't know if you are done.
>
>

Let's find a solution for the previews and I think it can be merged. ;D

emanuele45:
I dropped a couple of discontinued cache engines.

Frenzie:
Ah, I was wondering if any of them had anything to offer over e.g. opcache. I guess that answers the question.

Spuds:
The other ones support caching of user data, eg database query results, the built in opcache only caches the "compiled" scripts.  You can run some of these in parallel with the built in opcache as long as you turn of their opcache abilities.  Have done this with both Xcache and APCu

Frenzie:
I see, so it would be preferable to enable APCu in addition to opcache then. The SMF wiki is no help in this regard either, btw. Since APC is described as "opcode cache for PHP" I thought it was useless in 5.5+, and the description of APCu, "APC stripped of opcode caching" isn't helpful *at all*. The fact that it caches query results is kind of a big deal.

Spuds:
All the "cache" settings in ElkArte refer to its ability to set/fetch/clear "user data"  and really has nothing to do with any "opcache", at least in a high level sense.

The zend opcache is very efficient, better than APC or Xcache "opcache" capabilities (of course being built in has it benefits) so using it along side of a cache system that allows storing of "user data" brings the most benefit.  You could disable the built in zend stuff and just go with APC or Xcache as well.  The point is you should only have one opcache engine running and you should have a "user data" cache running (for sites that benefit from a cache)

APCu, since it has no opcache, as well as Xcache configured with its opcache disabled, are good "user data" storage options that ElkArte can benefit / make use of.  I think memcache is also just user data but I have not really used that so can't be sure.

Frenzie:
@Spuds I added your explanation to https://github.com/elkarte/Elkarte/wiki/Server-Settings

I'll also give some thought as to how to improve the way this is presented from within the admin interface itself. Something as simple as changing "ElkArte supports caching through the use of accelerators." to "ElkArte supports caching *of user data (such as query results)* through the use of accelerators." might do the trick.

Spuds:
Improving the wording is always a good thing :smile:  Thanks for adding this to the wiki !

PR #2504, opened 2016-04-23
A tiny one

emanuele45:
Just a couple of very smallish ones.


Spuds:
> PHP Fatal error:  Hacking attempt...<em>(setup.php-358)</em> in /var/www/sources/database/Db-abstract.class.php on line 333

emanuele45:
Should be fixed... unless it is just hidden.

PR #2511, opened 2016-04-25
And a few more tweaks / fixes 

Spuds:


themes/default/css/index.css
```diff
@@ -5383,6 +5383,7 @@ a.help .icon {
 /* This fixes a few outliers. */
 .icon-top {
 	vertical-align: middle;
+	padding-bottom: 3px;
```
Vekseid: I am putting in a proper fix for this soon, before you get even more carried away : p

Spuds:
Figured you would fix it, that's why its labeled as ``` /* This fixes a few outliers. */``` so they can be removed when needed.   Is this one of the not centered icon->svg issues?

Vekseid:
The only icon-top left is the popup window created by smc_Popup.prototype.show, which currently only uses a couple icons - namely the envelope icon, which is properly centered. So this is a css issue I haven't been able to properly resolve yet.

The menu icons that are off vertical center are because the icons aren't centered.I'd recenter them but I think we want to replace some of them, so leaving for now.

sources/subs/BBC/Autolink.class.php
```diff
@@ -149,7 +149,7 @@ public function hasPossibleLink()
 	public function parseLinks(&$data)
 	{
 		// Switch out quotes really quick because they can cause problems.
```
Frenzie: Problems? What kind of problems? =P

PR #2512, opened 2016-04-25
Some styling fixes

Vekseid:
1) Use the 'new' shade of green for newness. A bit less neon >_>

2) Removed shadows as requested, left in comments.

3) Fixing some styling things that you were starting to get carried away with. <_<

4) Menu button icon sizes - we should change this through the size of the buttons themselves (or the size of the default font). More than 1.5em means buttons with icons will be oversized compared to buttons without.

On the last point - do we want to switch to a full 16px 1em font size? I know many of my forum members would prefer that, and some of yours do as well apparently.

### DIFF
--- a/themes/default/css/_light/index_light.css
+++ b/themes/default/css/_light/index_light.css
@@ -1874,25 +1874,27 @@ div.bbc_footnotes .meaction {
 
 /* Helpers */
 
-.icon-shade, .chevricon, button.icon, button > .icon, a > .icon, a.icon, a.warnicon, a > .warnicon, a.noticon, a > .infoicon {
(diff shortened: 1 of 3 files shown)


Spuds:
> that you were starting to get carried away with

Go team ... LOL

Frenzie:
Bye bye filter and drop-shadows! I won't miss you! xD

> On the last point - do we want to switch to a full 16px 1em font size? I know many of my forum members would prefer that, and some of yours do as well apparently.

Go for it.

Vekseid:
Well it's @Spuds and @emanuele45 who make that call, really >_> Should probably have more discussion in an issue.

PR #2513, opened 2016-04-26
Installer Ignore param wasn't recoginzed.

eurich:


emanuele45:
http://www.elkarte.net/community/index.php?topic=2886.0

I tend to prefer brackets (less ambiguous), of course I'm not going to "fix" all the codebase at once, but I don't mind having or not the brackets.

Frenzie:
Brackets should stay and be (organically) added imho.

It's not just clarity but also the ease of quickly adding something (e.g for debug).

eurich:
re-added the brackets, I'm fine with both versions but we should standardize that at some point.

Frenzie:
Eh, I don't exactly mind it the other way or anything. Legibility-wise, I actually strongly prefer it Python-style, without braces and with meaningful whitespace, even though that sometimes comes with quick debugging annoyances of its own. But the simple fact is that in PHP (and JS, Java, and all other C-style languages), whitespace doesn't mean anything. What seems quick and elegant at first quickly becomes annoying at best (ugh, I have to add braces *again*?) and misleading at worst.

```
if (blabla)
  statement;
  statement2; //d'oh, it'll always fire

if (blabla)
  if (albalb)
    statement;
  statement; //d'oh, not part of first if!
```

PR #2515, opened 2016-04-26
* fixed encodings .. 

eurich:
thanks to @Vekseid for the hints, IE 9-11 doesn't like the not encoded versions.


Spuds:
If you still have this open, could you change ```stroke-width='200'``` to ```stroke-width='18em'``` on ```.i-board-new::before``` I did not know you could use em when I added back the beSocial icons

PR #2516, opened 2016-04-28
Fixing oops, templatelets for e-mail and online status

Vekseid:
Please review before merging.

1) The e-mail stuff straddles the line between 'bug fix' and 'unwanted additional exposure'

2) I wasn't sure where I should put the template bits, so I stuck them in index for now. Had difficulty parsing what you were doing.

### DIFF
--- a/sources/controllers/Memberlist.controller.php
+++ b/sources/controllers/Memberlist.controller.php
@@ -190,6 +190,7 @@ class Memberlist_Controller extends Action_Controller
 		);
 
 		$context['can_send_pm'] = allowedTo('pm_send');
+		$context['can_send_email'] = allowedTo('send_email_to_members');
 
 		// Build the memberlist button array.
 		if ($context['in_search'])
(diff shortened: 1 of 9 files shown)


emanuele45:
https://github.com/elkarte/Elkarte/pull/2516/files#diff-67f3afd2549112d5f8194b79dda5a762L130
if the one viewing the message is allowed to send emails, but the poster doesn't allow others to see his email, then an empty li will be rendered.
Considering template_msg_email is returning the link, I'd use something like:
```
$email = template_msg_email($message['id'], $message['member']);

if ($context['can_send_email'] && !empty($email))
{
    $poster_div .= '<li>' . $email . '</li>';
}
```

Vekseid:
It won't render an empty li, it will render a non-linked e-mail icon. May want to make a separate icon for that, but that's just graphical gloss.

Can do your method too, it just calls it out as being explicitly hidden.

PR #2524, opened 2016-04-30
Some more cleaup on the way to Beta

Spuds:


emanuele45:
3bfdaed41d49037837cc968d9b14016ba0507114 <= nope, Pm_Error_Exception was intended to be Pm_Error_Exception and not Elk_Exception. The reason is the "Pm_Exception" accepts specific parameters... the problem is that I forgot to add the file of the exception class... :dizzy_face: 

In my repo it looks like https://gist.github.com/emanuele45/0ef539d6543e4fe12579bc643935109f and goes into /sources/subs/Exception/

emanuele45:
Darn button... I missed a piece of message. xD

Can you revet the commit and add the file?
Or should I merge and do the revert/add myself?

Spuds:
I'll add the file and revert the *cough* fix.   A brand new class that already has depreciated parameters :cry:

emanuele45:
Spuds wrote on 01/05/16 01:16:
>
> I'll add the file and revert the /cough/ fix. A brand new class that
> already has depreciated parameters :cry:
>

I know the feeling! xD
Though it's a problem of js VS non-js.
The js version of the PM recipients sends the IDs, while the non-js
sends the names.
The "deprecation" is more related to the fact I'd favour the js-version,
deprecating the non-js. Though I wouldn't remove it "soon".

emanuele45:
Interesting: github markdown doesn't parse properly quotes coming from emails.
GitHub hax bugz!

PR #2527, opened 2016-05-01
Some continued work towards beta

Spuds:
Update the calls/use of detectServer()-> for set_time_limit and setTimeLimit and setMemoryLimit
Few more missing inits, some were global (bugs) others just nice to have
The $frequency -> $lang_data thing
Some random docs, parameter hint tweaks, & {}'s


emanuele45:
Re: 031d240f4aae203a9c8d6941bd1d1bb54085997f what would you think about make modSettings similar to the _req param?

I'm not sure what I'm talking about, but something like:
```php
abstract class AbstractModel
{
    protected $_db = null;
    protected $_modSettings = array();

    public function __construct($db, $modSettings)
    {
        $this->_db = $db;
        $this->_modSettings = new ArrayObject($modSettings);
    }
}

[...]

class Pbe_Imap extends AbstractModel
{
    public function __construct($db, $modSettings)
    {
        parent::__construct($db, $modSettings);

        $this->_hostname = $this->_modSettings->maillist_imap_host;
        $this->_username = $this->_modSettings->maillist_imap_uid;
        $this->_password = $this->_modSettings->maillist_imap_pass;
        [...]
    }
}

[...]

$pbe = new Pbe_Imap(null, $modSettings);
```

And then, while working, make most if not all the classes (except the controllers) extend the AbstractModel.
/me has no idea if it is good design or not...

Spuds:
Yup ... As I entered that change I was thinking along the same lines, be nice to have a globals getter ... could use the direct as you showed above as well as getSetting(var, default) as well.  Maybe something for B2 to investigate?

emanuele45:
Yep, the AbstractModel class is already there, I think I used it just once (i.e. when I had the idea to create it), if the idea seems right it can be explored, otherwise killed. :smile_cat:

Spuds:
Ah I see it ... used in mentions to bring in the db object ... OK I'll do a quick update to that and the imap thingy to use it as well in just a bit.  Got distracted by the draft issues (you one naughty boy)

PR #2549, opened 2016-05-21
Ajax for profile tabs

Spuds:
Ignore the minor things in the PR and look at 

https://github.com/elkarte/Elkarte/commit/f989f89b10c4393b5f05db49b22aec68fafd3705
https://github.com/elkarte/Elkarte/commit/950d85fdb157ea4a029c87abd280434c4b91df7a
https://github.com/elkarte/Elkarte/commit/f72e92f746ae6b914d2e7101d9b4691ad2c854ce
https://github.com/elkarte/Elkarte/commit/fa3c224329f0eb7d301301a92cea0a6b266a2a57

This is moving the profile Buddies tab to be ajax.   Have not looked at "recent" tab yet but that would be next. 

To make this work UI tabs allows us to specify an href value to load ajax html into a tab when its activated.  Given the way tabs work, it has to be HTML and not XML data (its called to late in the stack to effect the ajax settings).  So is kind of "ugly" having the action inside of xml.controller since it does not return XML.  You could use the tab load event to trigger a AJAX / XML event and then build the values but that IMO gets super ugly in the JS, at least any way I could see this working.

Had to do a minor refactor of ProfileInfo.controller so that one could call the buddies function from the xml controller (or anywhere), as to much user context was built inside of the main method. So this just broke it up in to small methods so one can call the needed methods to get things rolling.

So take a look, give me thoughts on if we should proceed or make change or leave things as they were.


emanuele45:
It's 1:00 AM here so I'm a little sleepy and may be asking questions already answered.

Why the echo in the controller instead of use a sub_template?
Something similar to what I did for the poll-create interface ([the js](https://github.com/elkarte/Elkarte/blob/bc0d8a640e9cc51bdc0e8c19a6d75e3c0cb5a414/themes/default/scripts/script_elk.js#L1395), [the php](https://github.com/elkarte/Elkarte/blob/bc0d8a640e9cc51bdc0e8c19a6d75e3c0cb5a414/sources/controllers/Poll.controller.php#L742) and [the template](https://github.com/elkarte/Elkarte/blob/bc0d8a640e9cc51bdc0e8c19a6d75e3c0cb5a414/themes/default/Poll.template.php#L19))?

> You could use the tab load event to trigger a AJAX / XML event and then build the values but that IMO gets super ugly in the JS, at least any way I could see this working.

If I got that one right, I agree. It would require something like angular.js to look somehow nice...

Spuds:
Hummm ... I think its a left over when I was messing around with an XML response and a JS loop to build the inner div .. it sorta worked but you had to jsvar all of the $context stuff required by buddies and those sub templates etc .. just got so messy, at least as I approached it.  

Anyway yeah the tempalte could / should echo as normal

themes/default/languages/english/Profile.english.php
```diff
@@ -419,7 +419,7 @@
 $txt['profile_warning_effect_watch'] = 'User will be added to moderator watch list.';
 $txt['profile_warning_effect_own_watched'] = 'You are on the moderator watch list.';
 $txt['profile_warning_is_watch'] = 'being watched';
-$txt['profile_warning_effect_moderation'] = 'All users posts will be moderated.';
+$txt['profile_warning_effect_moderate'] = 'All users posts will be moderated.';
```
Frenzie: Should be either `user` or `user\'s`. But if I catch the meaning correctly I'd better be rewritten to something like `All posts by this user will be moderated.`

Spuds:
These last few commits added the recent tab as ajax loaded, along with the output remaining in the templates, etc.

Spuds:
And since having them in xml controller was fugly, I updated ProfileInfo_Controller to be able to dispath those actions.  So this should be the *cough* end of this :innocent:

emanuele45:
Yup, makes sense.
What about the obExit?

Spuds:
> What about the obExit?

Suggestions?

emanuele45:
Something like this?
https://github.com/Spuds/Elkarte/compare/mybugs...emanuele45:mybugs
Well, without the debug code of course. :P 

I'm not entirely sure (yet) why the header is necessary (since the poll interface is loaded without it, but most likely is because of something broken in the polls).

Spuds:
Did that work? 

I'll certainly give it a try but I thought ;xml would output an xml header.  The jquery ajax object while available in the tabs widget, has been mostly processed (type set, data set, etc) so you can't tell it that its XML data, nor would I want to write the code to process all of that array data anyway.

emanuele45:
Yep, locally it worked (oh BTW I also had to change ```action=profileinfo``` to ```action=profileInfo``` (upper-case "I") because of Linux sensitivity to cases lol).

Yep about xml, that's why I had to force a header in the controller to ensure it is html.

Spuds:
Well I committed the "form over function" :smile: changes to this branch but I'll be dammed if I know where they are!   Looks git github has some problem going on, my client says everything is up to date but the commit does not show up here ... :anger:

live627:
It's there, I see it...

Spuds:
I noticed on Gits status page yesterday

> Some users may experience a delay in pushes or other changes appearing on the site.

It did take a couple of hours but it finally did show up.

PR #2551, opened 2016-05-22
Editor, don't grow too tall when there is an error

live627:


emanuele45:
Removing those two, does the editor (still) resizes while changing the size of the screen?
I remember I added them for some reason, but honestly I don't remember which one... :blush:

PR #2554, opened 2016-05-26
Pedestrian changes

Spuds:
spelling, docs, newlines, return points and the like as we prep for formal B1


sources/Debug.class.php
```diff
@@ -174,7 +174,7 @@ public function merge_db($value)
 	public function get_db()
 	{
 		if (!$this->_track)
-			return;
+			return '';
```
Frenzie: Would FALSE be better?

Spuds: possibility ... ```$_SESSION['debug_redirect'] = Debug::get()->get_db();```

Its value is checked for empty, so '', null, false or even array() would be fine.   Really array() maybe the most correct as I think thats what that value holds when in use.  Technically you can just use return; and it will return null back, I just don't like that since the calling function should expect something back  :innocent:

PR #2556, opened 2016-05-27
Inline JS minification

Spuds:
This is a bit of an experiment. 

It allows us to use the "internal" squeeze to minify the standard and deferred Inline javascript that templates, sources, addons, etc add to the page.  This just adds (really separates and exposes) the code in Combiner that was available to do this.  Also adds a getInstance method to Combiner since we were invoking it a few times.

The plus side is that it will save a 1-2k per page load, the downside is that it will also add a few hundreds of a second to the page creation time.  I considered caching the results, but could not see a suitable "key" since whats often inline is unique to the user, and perhaps unique on each page load.

Also updated to the latest squeeze.

### DIFF
--- a/sources/SiteCombiner.class.php
+++ b/sources/SiteCombiner.class.php
@@ -24,6 +24,11 @@ if (!defined('ELK'))
 class Site_Combiner
 {
 	/**
+	 * Instance of the class
+	 */
+	private static $_instance = null;
+
+	/**
 	 * Holds all the files contents that we have joined in to one
 	 *
 	 * @var array
@@ -241,6 +246,23 @@ class Site_Combiner
 	}
 
 	/**
+	 * Compress a supplied JS string for output in the template
+	 *
+	 * @param string $string
+	 *
+	 * @return string
+	 */
+	public function inline_js_compress($string)
+	{
+		$min_string = $this->_squeeze_me($string);
+
+		if ($min_string === false || trim($min_string) === '')
+			$min_string = $string;
+
+		return $min_string;
+	}
+
+	/**
 	 * Add all the file parameters to the $_combine_files array
 	 *
 	 * What it does:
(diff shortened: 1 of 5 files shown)


Frenzie:
Wouldn't it be much more worthwhile to get rid of the inline stuff and to add it dynamically on DOMContentLoaded or some such? That would surely save at least as much (browser caching!) without any server issues.

Spuds:
I suppose, but I have no real idea how to do that :/ ... You should open a feature request for this or better make a PR, there is a lot of source generated JS that is loaded to the page.   I'll close this one as not needed and await a better way to handle this.

Frenzie:
Mostly it's just that I don't see 1-2 kB (is that before or after gzip?) as a worthwhile trade-off for hundreds of a second of extra processing time. But I actually misinterpreted this while I was distracted by the fact that my heater/boiler broke this morning. By default there seems to be about 2 kB of JS which is virtually incompressible through this method in the header and footer (I mean all the stuff like `likemsg_are_you_sure = 'You already liked this message, are you sure you want to remove your like?',`), while I was actually thinking of stuff that's just echoed and wouldn't be affected. Overall I just don't quite see how you'd get more than a couple hundred bytes out it (mostly by stripping whitespace) that you wouldn't also get with gzip?

Spuds:
Bummer about the heater ... but its summer now, you have months to get that fixed :dancer: 

1-2k was a before vs after final page size, I don't recall which pages ATM, so probably more than just white space *shrugs*  Some pages have quite a bit of JS created in the sources and templates (some of which is erroneously not deferred), addons often add quite a lot.

No idea of the impact in the gzip size of the transport (will vary on several factors anyway such as web server used or php buffer based or none at all), nor cpu / time impact of the zip/unzip for the two base page sizes.

Frenzie:
It's the shower that matters. ;)

emanuele45:
Cold shower helps! :P

PR #2569, opened 2016-06-01
A bunch of tests

emanuele45:
Some more tests for our beloved BBCode parser! :dancers:


Spuds:
I'm assuming the B2 milestone means wait to merge.

emanuele45:
I'm adding some more, but the B2 means... no idea. xD

emanuele45:
And that's all for now.
The only missing is quote because it has too many variants... :confounded: 

@joshuaadickerson FYI :stuck_out_tongue:

joshuaadickerson:
That's not all of them, right?

joshuaadickerson:
You can set tests to fail and continue. That would be better than allowing the bug. It should bug us ;)

emanuele45:
The only not tested (yet) are the various forms of the quote tag.

emanuele45:
YAY!
Added (most of) the quote variants too.

PR #2571, opened 2016-06-05
A couple fo tweaks to the D&D area

emanuele45:


emanuele45:
This has become experimental now! :rotating_light:

Apart from e384c7521ab36ef02cc455f39d372495df961e16 that slipped in, but is necessary anyway, I added a few commits with some experiments at using the "semantic" html5 tags in the three main templates (boardindex, messageindex and display... plus of course index).
I took inspiration from wordpress.
It's nothing too fancy, just a few that seemed somehow logical.

Comments are welcome!

Spuds:
Well if its good enough for wordpress :tongue: 

I have no idea on whats correct for those areas but I agree trying to use them as appropriate is a good thing to do but I would not get in to a brain freeze deciding if you should use div vs section etc.

emanuele45:
I guess nobody knows what is "correct" (I think there is no such a thing as "correct"), so I applied my own common sense... if that is wrong, who cares. :stuck_out_tongue: 
The idea I applied is that there are few pieces of markup that **are** something that is not relevant content, so my intention was to give more evidence to the "real" content of the page changing the div to something the seemed good enough (e.g. header for the... header, footer for the footer, article for the posts, etc.).
I dropped in some other fancy tag just for the sake of it (and because to me it seemed not a wrong choice).

PR #2596, opened 2016-06-20
Rewire index

emanuele45:
Opinions welcome.
I turned upside-down part of index.php and the dispatcher.

The idea behind is that it's easier to ask the controller if he needs the theme or to track the stats rather than having to use an hook to create an array of values and do in_array, etc.
To me it looks cleaner and even more readable now.

Reminder: if the PR is not accepted, there is a bug in the ```if``` to check if dlattach should not load the whole theme.


live627:
I  like this. Much more clean and logical.

emanuele45:
darn... conflicts... :sob:

emanuele45:
Well, judging by the thumbs-up I feel the PR makes sense.
Before merging it I'd stress this PR removes [one hook](https://github.com/elkarte/Elkarte/pull/2596/commits/dd2f84bf6786cd6e940992c7b2af278bcf9b3858#diff-828e0013b8f3bc1bb22b4f57172b019dL247)  (because its function is now taken care by the controller itself) and [removes one parameter](https://github.com/elkarte/Elkarte/pull/2596/commits/c018098f726c514e07e74625189f4e97a17da344#diff-28350c611aca9e83f41fc0ec474e1f1eL191) from another one (because the directory is now taken care by the autoloader).

If there are no objection in a couple of days I'll merge. :angel:

live627:
>  this PR removes one hook 

I wonder if anyone other than portals used it.

emanuele45:
That's my feeling as well. I had a moment thinking about adding a way to keep supporting the hook at least for 1.1, but then I felt it may be easier and better support anyone in need of help upgrading custom code.

Spuds:
I used it in SP but no where else .. it will be a lot nicer moving it to the controller.

live627:
My addon just reacted to this:
```
        // Hey man, what happened? /stoner voice
```

To add some context: my main controller used its constructor to initialise things which in turn broke upon this update.

Bleeding edge development sure is fun! :P

Spuds:
Dave's not here, man :P

Issue #2599, opened 2016-06-21
Update markdown library?


emanuele45:
I noticed we are shipping the 3 years old version of the markdown library, while there is a branch (lib) actively developed that reached version 1.6 last December:
https://github.com/michelf/php-markdown/branches

Time to update it?

KeiroD:
Definitely time to update it. :P

Spuds:
We should .. can't use it in the 1.0 branch but its fine for 1.1 .. removed the question label

emanuele45:
This was merged.

PR #2600, opened 2016-06-21
Php 7

emanuele45:
A set of changes to make Elk a bit more php-7-ready.

I have just few doubts:
  1. substr() now returns an empty string, if string is equal to start characters long. 
  2. the eval in Cache.class.php (I suspect it should be extracted from the if and put in a try/catch)
  3. the list comment <blockquote>In general, it is recommended not to rely on the order in which list() assignments occur, as this is an implementation detail that may change again in the future. </blockquote>
  4. #2599 
  5. the changes to foreach may create some subtle bug http://php.net/manual/en/migration70.incompatible.php#migration70.incompatible.foreach but review all the foreach is basically impossible, so let's hope to find them during beta (yeah, I like betting! :stuck_out_tongue_winking_eye:  )

and that's all.

### DIFF
--- a/sources/Debug.class.php
+++ b/sources/Debug.class.php
@@ -181,7 +181,7 @@ class Debug
 
 	/**
 	 * Adds a new getrusage value (by default two are added: one at the beginning
-	 * of the script execution and one at the end
+	 * of the script execution and one at the end)
 	 *
 	 * @param string $point can be end or start depending on when the function
 	 *               is called
@@ -189,6 +189,7 @@ class Debug
 	 */
 	public function rusage($point, $rusage = null)
 	{
+		// getrusage is missing in php < 7 on Windows
 		if (!function_exists('getrusage') || !$this->_track)
 			return;
 
--- a/sources/Debug.class.php
+++ b/sources/Debug.class.php
@@ -190,8 +190,10 @@ class Debug
 	public function rusage($point, $rusage = null)
 	{
 		// getrusage is missing in php < 7 on Windows
-		if (!function_exists('getrusage') || !$this->_track)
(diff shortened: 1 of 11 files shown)


emanuele45:
I'm not sure if it is worth backporting these to 1.0.
If 1.1 development goes smooth enough, we may release 1.1 well before the end of the year. In that case I wouldn't bother.

sources/Debug.class.php
```diff
@@ -181,14 +181,15 @@ public function get_db()
...
+	 * of the script execution and one at the end)
 	 *
 	 * @param string $point can be end or start depending on when the function
 	 *               is called
 	 * @param string[]|null $rusage value of getrusage or null to let the method call it
 	 */
 	public function rusage($point, $rusage = null)
 	{
+		// getrusage is missing in php < 7 on Windows
 		if (!function_exists('getrusage') || !$this->_track)
```
Spuds: NitPik I'd reverse this order .. no need to do the function exists if you are not tracking to begin with.

Spuds:
I suppose its fine to use this in 1.0.8 ( I don';t see any harm ), but until we know more we can't claim php7 compatibility.  Some of the changes like list(,,$bla) no longer working and the foreach by value change description reads really oddly.

emanuele45:
Yep, yep indeed.

sources/Debug.class.php
emanuele45: done!

live627:
> foreach by value change description reads really oddly.

Eh?

emanuele45:
Well, I guess it can be merged. :angel:

PR #2603, opened 2016-06-26
Update DataValidator.class.php

live627:
Use filter_var() so that we can test form input against booleans, integers, and floats

### DIFF
--- a/sources/subs/DataValidator.class.php
+++ b/sources/subs/DataValidator.class.php
@@ -986,7 +986,7 @@ class Data_Validator
 		if (!isset($input[$field]))
 			return;
 
-		if (!is_int($input[$field]))
+		if (filter_var($input[$field], FILTER_VALIDATE_INT) === false)
 		{
 			return array(
 				'field' => $field,
@@ -1011,7 +1011,7 @@ class Data_Validator
 		if (!isset($input[$field]))
 			return;
 
-		if (!is_bool($input[$field]))
+		if (filter_var($input[$field], FILTER_VALIDATE_BOOLEAN, FILTER_NULL_ON_FAILURE) === null)
 		{
 			return array(
 				'field' => $field,
@@ -1036,7 +1036,7 @@ class Data_Validator
 		if (!isset($input[$field]))
 			return;
 
-		if (!is_float($input[$field]))
+		if (filter_var($input[$field], FILTER_VALIDATE_FLOAT) === false)
 		{
 			return array(
 				'field' => $field,
@@ -1272,7 +1272,7 @@ class Data_Validator
 	}
 
 	/**
-	 * Uses Util::htmlspecialchars to sanitize any html in the input
+	 * Uses Util::htmlspecialchars to sanitize any html in the intput
 	 *
 	 * @param string $input
 	 * @param string|null $sanitation_parameters
--- a/sources/subs/DataValidator.class.php
+++ b/sources/subs/DataValidator.class.php
@@ -1272,7 +1272,7 @@ class Data_Validator
 	}
 
 	/**
-	 * Uses Util::htmlspecialchars to sanitize any html in the intput
+	 * Uses Util::htmlspecialchars to sanitize any html in the input
 	 *
 	 * @param string $input
 	 * @param string|null $sanitation_parameters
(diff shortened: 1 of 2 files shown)


live627:
> TestDataValidator::testValidation
Test: required failed data: :D but it should have passed
Failed asserting that Array &0 () is null.

This is only on PHP 5.3. I don't know why.

emanuele45:
Okay, there are two problems here:
1. php 5.3 fails on testing booleans, and that is the reason the test fails,
2. the validation class picks the first error in case of validation of arrays, but that error may refers to any element of the array, making it a bit more difficult to identify the issue.

For 1 I have no idea, so it's up to you fix it. :stuck_out_tongue: 
For 2 I have a couple of ideas, but better open an issue and discuss there, also because I feel I need @Spuds input. :stuck_out_tongue:

live627:
1. fixed
2. not relevant here

emanuele45:
Though, false is a proper boolean AFAIK, so it should "validate" properly... am I wrong?

live627:
https://github.com/live627/Elkarte/blob/d039c696fb8934f3c9d599890974ce3673fb8275/tests/sources/subs/DataValidator.class.Test.php#L145

emanuele45:
Let me try to recap to be sure I understood: at the moment, filter_var in PHP < 5.4 returns "null" if a valid "false" is passed, and that means "error" to the Data_Validator class.
Right?

If what I wrote is correct, then the fix is not to change the test to true, this is just hiding the problem (that is a valid false will not pass the test of the class). The fix is make the Data_Validator class recognize a false as a boolean consistently across all the supported PHP versions (i.e 5.3 and up).

Considering that FILTER_NULL_ON_FAILURE has to stay, the only solution is use is_bool for PHP < 5.4, but at that point I wonder is worth maintaining the two versions of the test.

live627:
> Let me try to recap to be sure I understood: at the moment, filter_var in PHP < 5.4 returns "null" if a valid "false" is passed, and that means "error" to the Data_Validator class.

Correct.

> I wonder is worth maintaining the two versions of the test

Probably best to split all of those tests in the big arrays to each their own methods.

emanuele45:
>     I wonder is worth maintaining the two versions of the test
>
> Probably best to split all of those tests in the big arrays to each
> their own methods.
>

Sorry, poor wording from my side...
I was thinking that in Data_Validator::_validate_boolean there should be
something like:
```
if (version_compare(PHP_VERSION, 5.4, '<'))
{
    // use is_bool
}
else
{
    // use filter_var
}
```

because the main point of wrapping php functions into our own classes is
to clean up within the class all the php inconsistencies across
versions/operating systems/etc. otherwise it becomes just a wrapper.

So my "two versions of the test" was meant to mean the is_bool and the
filter_var together in _validate_boolean.

live627:
Well, let's back up for a moment. What is the primary use for the data validator? We need more usage data, but I foresee it being used against form submissions. That's how I use it, at least. And straight up booleans are not going to be sent, but strings, which validate fine anyway.

I dunno, I just think that it's a waste to officially support what I believe to be a fringe case on an old platform [that's been unsupported for nearly two years](http://php.net/eol.php)

Spuds:
I don't believe the boolean check has been used in the code to date.   live627 has a very valid point, how would the check be used to begin with?  The original intention was to validate in a common way all of the cruft that would be submitted by forms.

As the code was, with ```is_bool```, only TRUE and FALSE would be valid, actual bool's.  

filter_var's ```FILTER_VALIDATE_BOOLEAN``` extends that to be would php act on this value as a bool.  As such  1, '1', 0, '0', true false, 'on' 'off', '' etc would be considered as valid true's and false's. At least thats how I interpret the docs.

Perhaps we should leave the boolean check as it was, and add a new yesno check that utilizes filter_var capabilities.  Or if just the extended boolean check would the double check work instead of the null.
```if (filter_var($input[$field], FILTER_VALIDATE_BOOLEAN) === true) || (filter_var($input[$field], FILTER_VALIDATE_BOOLEAN, FILTER_NULL_ON_FAILURE) === false)```

Could also break that up to ison and isoff checks, really depends on how one would interact with form data.

**Comment by emanuele45** [comment]
The only thing I feel is important is consistency: if (by sheer luck) I have to validate a boolean, for example because
```
$_POST['form_data'] = json_encode(array('a' => false));
$b = json_decode($_POST['form_data'], true);
```
(yes, the first line is made up, but in my mind it could be the result of POST'ing a json_encode'd array through some AJAX call), I would hope the class returns the exact same result across the various PHP versions we support, and not have to implement some ugly:
```
if ($validator->get_errors() || (PHP_VER < 5.4 && variable === false))
{
    // do the job
}
```

I would prefer to hide this complexity inside the class so that I don't have to care in the future, but that's just me. :)

emanuele45:
I still think the class should behave the same no matter the various versions of php, but I have time to change it later. :stuck_out_tongue:

PR #2608, opened 2016-06-27
Use SSL when sending postback to PayPal

live627:
Question: Keep using the `www` endpoint or switch to the recommended `ipnpb`?


Spuds:
Having not, thankfully, read the IPN guide in quite some time, I have no real idea on the consequences of the two.  

Generally if something is recommended on API's then I try to use it.  I'd guess using the ```www``` redirects to ```ipnpb``` so it would seem logical to move to that endpoint.

emanuele45:
:+1: to moving to ipnpb.

Should 7c2ad036bb32dc9d3bf0f720cb2d6a10bd45eec3 use ssl instead of https as well?

live627:
> Should 7c2ad03 use ssl instead of https as well?

Already uses it.

PR #2626, opened 2016-06-29
Call the frontpage method directly

live627:
This removes integrate_action_frontpage since I think it is overkill. This can prevent conflicts with addons that don't understand the system well enough.


emanuele45:
I guess it makes sense cutting the middle-man.

PR #2628, opened 2016-06-30
Allow forks to initiate their own builds

live627:
repositories enabled in 2015 or later must have this line

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: true
 language: php
 
 matrix:


emanuele45:
So this is needed in 1.0 as well, right?

live627:
No. That branch seems to not need it as evidenced by https://travis-ci.org/live627/Elkarte/builds/140444469

emanuele45:
Interesting... oh well.

PR #2630, opened 2016-07-01
More cleanup of the site dispatcher

live627:
This is my go at cleaning up one of the most important files in Elkarte.

A fool's work is never done, though.

### DIFF
--- a/sources/SiteDispatcher.class.php
+++ b/sources/SiteDispatcher.class.php
@@ -324,36 +324,6 @@ class Site_Dispatcher
 	}
 
 	/**
-	 * Backward compatibility function.
-	 * Determine the current action from $_GET
-	 * @deprecated since 1.1
-	 */
-	protected function _getAction()
-	{
-		return isset($_GET['action']) ? $_GET['action'] : '';
-	}
-
-	/**
-	 * Backward compatibility function.
-	 * Determine the current subaction from $_GET
-	 * @deprecated since 1.1
-	 */
-	protected function _getSubAction()
-	{
-		return isset($_GET['sa']) ? $_GET['sa'] : '';
-	}
-
-	/**
-	 * Backward compatibility function.
-	 * Determine the current area from $_GET
-	 * @deprecated since 1.1
-	 */
-	protected function _getArea()
-	{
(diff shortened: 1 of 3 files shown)


emanuele45:
Juts a technical remark: it would be better to avoid merges, but use rebase instead, if you have problems with it, I can do the rebase for you, I got quite used to them. lol

sources/SiteDispatcher.class.php
```diff
@@ -88,6 +92,11 @@ public function __construct($action = null, $subaction = null, $area = null)
 		call_integration_hook('integrate_action_frontpage', array(&$this->_default_action));
 
 		$this->_noActionActions($action, !empty($modSettings['allow_guestAccess']));
```
Spuds: $action is not defined anylonger

sources/SiteDispatcher.class.php
```diff
@@ -88,6 +92,11 @@ public function __construct($action = null, $subaction = null, $area = null)
...
+		// A safer way to work with our form globals
+		$_req = HttpReq::instance();
```
Spuds: With doing this, do we need to pass it ($-req) to the constructor any longer?

emanuele45:
Apart from that I'd say... cool stuff! :dancer:

Spuds:
LOL .. I made comments right while live627 was making the updates to the commit.

live627:
> LOL .. I made comments right while live627 was making the updates to the commit.

Yeah... bad merge.

> it would be better to avoid merges, but use rebase instead

Why? To get rid of merge commits?

emanuele45:
Sorry, I got the email notification only today... :-\ 

Yep, that and because the tree becomes cleaner.

live627:
GH has the option to not make a merge commit in the repo options-I just now found this out accidentally.

emanuele45:
GH has an option to squash all the commits in a PR into a single commit (squash and merge), what I was suggesting is a little different (rebase instead of the merge you did to keep up with the changes in development).
Anyway, this is already here, and we have done it in the past, so I guess I can avoid being too picky.

PR #2634, opened 2016-07-04
Set the display cache for the custom profile fields

### LINKED ISSUE #2479: Custom fields don't show up in topic view until explicitly saved in Profile Fields
1. Enable *Advanced Profile Fields* in *Core Features*
2. The default fields are now working (i.e. available to edit in *Forum Profile*)
3. Add some info into the fields in *Forum Profile*
4. The icons don't show up in the profile info dropdown until after you click *Save* (not changing a thing about the defaults) or modify a setting in _Configuration_ → _Features and Options_ → _Profile Fields_

This applies to both 1.0.7 and the 1.1 dev branch.

Edited to clarify.

live627:
Fixes #2479

### DIFF
--- a/sources/subs/Admin.subs.php
+++ b/sources/subs/Admin.subs.php
@@ -478,10 +478,17 @@ function custom_profiles_toggle_callback($value)
 
 	if (!$value)
 	{
+		// Disable all fields. Wouldn't want any to show when the feature is disabled.
 		$db->query('', '
 			UPDATE {db_prefix}custom_fields
 			SET active = 0'
 		);
+	}
+	else
+	{
+		// Set the display cache for the custom profile fields.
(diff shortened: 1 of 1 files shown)


emanuele45:
Ohhh... so it's a cache thing!

If so, it would be better to have it in setting_callback (or maybe in on_save) when enabling/disabling the core feature, wouldn't it?
Also, since this applies to 1.0.x as well, if you don't mind, could you send it to the 1.0.8 patch, so that we will get the fix here as well when the patch is merged. :innocent:

live627:
Sounds like a plan!

live627:
Changes made

emanuele45:
Perfect!
I grabbed the other one, the fix will come in with the merging of the patch branch to development in... a few days, I was hoping to release sooner, but I get lost in some other stuff. :angel:

PR #2636, opened 2016-07-06
Set the display cache for the custom profile fields

live627:
Backport #2634


live627:
This one is signed off because of the invalid author.

emanuele45:
Thanks! :D

PR #2638, opened 2016-07-29
Fix session handler, as Db::insert returns nothing.

kode54:
Maybe the Database classes should be changed instead? That's a big assumption to expect code to not check a return value from insert(), since it may be expected to return success or failure.

### DIFF
--- a/sources/Session.php
+++ b/sources/Session.php
@@ -188,7 +188,7 @@ function sessionWrite($session_id, $data)
 	$db = database();
 
 	// First try to update an existing row...
-	$result = $db->query('', '
+	$db->query('', '
 		UPDATE {db_prefix}sessions
 		SET data = {string:data}, last_update = {int:last_update}
 		WHERE session_id = {string:session_id}',
@@ -199,8 +199,10 @@ function sessionWrite($session_id, $data)
 		)
 	);
 
+	$result = $db->affected_rows();
+
 	// If that didn't work, try inserting a new one.
-	if ($result === false)
+	if (empty($result))
 	{
 		$db->insert('ignore',
 			'{db_prefix}sessions',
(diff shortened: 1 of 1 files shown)


emanuele45:
Something seems broken...

kode54:
This is what I get for making a PR against a patch branch instead of master.

kode54:
Bumping to post a comment about why checking the row count for the update query is a bad idea. At least with the version of MariaDB I have installed, if you update a row with the same values it currently contains, it will return 0 rows affected. Sure, this shouldn't happen if the timestamp is different, but what if multiple session writes occur in the same second?

kode54:
Checking for dupes now, rolling another update.

emanuele45:
> This is what I get for making a PR against a patch branch instead of master.

Well, that's the correct thing to do. :wink: 

Let's try to get the conditions the error occur: what version of php do you have? What version of mariadb?

kode54:
PHP 7.0.8
MariaDB 5.5.47

The error occurs if someone triggers the session save multiple times in one second. The logged errors all occur grouped into one timestamp, 2-4 errors per instance.

emanuele45:
So the issue is that when a session is stored with the same timestamp and the same data...
Could you try adding in Session.php after the line:
```
$sc = $_SESSION['session_value'];
```
this:
```
$_SESSION['mictrotime'] = microtime();
```

it's mainly a workaround, but it should change the data and force the db to return a affected_rows in all the cases (unless of very unfortunate events).

... unless I'm wrong, of course! lol

kode54:
That seems to work. I'll let you know if any further session issues arise.

emanuele45:
I submitted the workaround discussed above in the 1.0.8.1 path.
I'm closing this one so that we don't have PR hanging, probably we'd need a better fix, in that case, feel free to open a new issue referencing this PR for the discussion. ;D

albertlast:
About this mysql issue (is not a php7 issue in my eyes),
i fixed in this way: https://github.com/SimpleMachines/SMF2.1/pull/3545

emanuele45:
This morning I did a search and I remember there is just one situation that may be problematic in respect to this change... but I forgot to take not and I forgot which one it is... :sob:

PR #2659, opened 2016-08-19
Show profile errors

live627:
Fix highlighting of fields with errors.

The entire profile area is a festering mess.


sources/subs/Profile.subs.php
```diff
@@ -114,7 +114,7 @@ function setupProfileContext($fields, $hook = '')
 			}
 
 			// Was there an error with this field on posting?
-			if (isset($context['profile_errors'][$field]))
+			if (isset($context['post_errors '][$field]))
```
emanuele45: There is a space at the end of "post_errors", before the closing single quote. ;)

live627: fixed

PR #2663, opened 2016-08-21
Inline permssions and some work on permissions

emanuele45:
This is for B3, here just as reminder and to get opinions (if any :stuck_out_tongue: ).


sources/subs/InlinePermissionsForm.class.php
```diff
@@ -0,0 +1,180 @@
...
+		);
+
+		// ...and replace them with new ones.
+		if (!empty($insertRows))
+			$db->insert('insert',
+				'{db_prefix}permissions',
+				array('id_group' => 'int', 'permission' => 'string', 'add_deny' => 'int'),
+				$insertRows,
+				array('id_group', 'permission')
+			);
```
live627: Why not use the ready made function for this?

		$permChange = array();
		foreach ($permissions as $permission)
		{
			if (!isset($_POST[$permission]))
				continue;

			foreach ($_POST[$permission] as $id_group => $value)
			{
				if (in_array($value, array('on', 'deny')) && (empty($context['illegal_permissions']) || !in_array($permission, $context['illegal_permissions'])))
					$permChange[] = array($permission, (int) $id_group, $value == 'on' ? 1 : 0);
			}
		}

		require_once(SUBSDIR . '/ManagePermissions.subs.php');
		replacePermission($permChange);

emanuele45: Good point, this should do https://github.com/elkarte/Elkarte/pull/2663/commits/91fd9737ede039399a234602fbe2e4631c9a6dd1 (I hope)

sources/subs/InlinePermissionsForm.class.php
```diff
@@ -47,7 +47,7 @@ public static function save($permissions)
 			foreach ($_POST[$permission] as $id_group => $value)
 			{
 				if (in_array($value, array('on', 'deny')) && (empty($context['illegal_permissions']) || !in_array($permission, $context['illegal_permissions'])))
-					$insertRows[] = array((int) $id_group, $permission, $value == 'on' ? 1 : 0);
+					$insertRows[] = array('id_group' => (int) $id_group, 'permission' => $permission, 'add_deny' => $value == 'on' ? 1 : 0);
```
live627: The parameters now need reordered

		$insertRows[] = array($permission, (int) $id_group, $value == 'on' ? 1 : 0);

Array keys are ignored.

sources/subs/InlinePermissionsForm.class.php
```diff
@@ -0,0 +1,178 @@
...
+
+			foreach ($_POST[$permission] as $id_group => $value)
+			{
+				if (in_array($value, array('on', 'deny')) && (empty($context['illegal_permissions']) || !in_array($permission, $context['illegal_permissions'])))
+					$insertRows[] = array('id_group' => (int) $id_group, 'permission' => $permission, 'add_deny' => $value == 'on' ? 1 : 0);
+			}
+		}
+
+		// Remove the old permissions...
+		$db->query('', '
```
live627: This query is now redundant.

sources/subs/Permissions.class.php
```diff
@@ -0,0 +1,201 @@
...
+			// Do the insert.
+			if (!empty($permissions))
+			{
+				$db->insert('insert',
+					'{db_prefix}board_permissions',
+					array('id_group' => 'int', 'id_profile' => 'int', 'permission' => 'string', 'add_deny' => 'int'),
+					$permissions,
+					array('id_group', 'id_profile', 'permission')
+				);
+			}
```
live627: Another one.

			$permissions = array();
			while ($row = $db->fetch_assoc($request))
				foreach ($children[$row['id_group']] as $child)
					$permissions[] = array($row['permission'], $child, $row['add_deny'], $row['id_profile']);
			$db->free_result($request);

			// Do the insert.
			if (!empty($permissions))
			{
				replaceBoardPermission($permissions);
			}

sources/subs/InlinePermissionsForm.class.php
emanuele45: If they do it's a bug:
https://github.com/elkarte/Elkarte/blob/development/sources/database/Db-abstract.class.php#L270-L281
I'll do some tests later.

sources/subs/InlinePermissionsForm.class.php
emanuele45: Why is redundant? I didn't see it in the function.
https://github.com/elkarte/Elkarte/blob/d210298c7a53f0ed94ecd04b291f4ab0240528f6/sources/subs/ManagePermissions.subs.php#L964

sources/subs/Permissions.class.php
emanuele45: Good catch!

sources/subs/InlinePermissionsForm.class.php
live627: it replaces, so a delete is not needed.

live627: ... and this suggestion breaks turning the permission off. D'OH!

sources/subs/InlinePermissionsForm.class.php
live627: I was wrong. What can I say-that syntax is new to me.

emanuele45: I sneaked it in a while ago because I was tired of having to deal with inserts ordering! :grin:

live627: nice!

sources/subs/InlinePermissionsForm.class.php
emanuele45: I guess at some point the two queries may be put in the same function... later.

live627:
I'm continuing this branch https://github.com/elkarte/Elkarte/compare/development...live627:inline_perms

emanuele45:
Okay, I'll keep this open to track the "issue". ;)

live627:
Currently, Inline_Permissions_Form::save() checks the session, a token, and a permission. I'd argue that those belong in the controller. Also, they inhibit testing.

I'm going to remove them.

emanuele45:
**I think** (I forgot by now lol) my "long term" idea was to have:
1. a class that does the "form" (i.e. including security checks for session/tokens and **maybe** permissions, but this may really belong to the controller),
2. some other code that deals with changes in the db etc..
This was the first step of extracting everything from where it was and throw it in a single class (that would become the point 1.
The second step was to extract the changes and move them to somewhere else.

live627:
I really like the idea of a universal form object.  A while ago I made a form class for my own needs. I can re-purpose it for use in Elk.

https://gist.github.com/live627/6c35c6a23d647818d9c766947c400df7

emanuele45:
The idea of a form class comes out every so often, but has never been developed.
I think it would be pretty useful indeed. :smile_cat:

PR #2671, opened 2016-09-05
Small bunch of fixes

emanuele45:


Spuds:
Probably just me, but the levelHigher levelLower methods still need better names.  levelHigher(3) is true for 3 ... which is not higher but equal, etc.  I get the intent, but since it is for readability, perhaps a better name still ?

emanuele45:
Yeah... I had similar thoughts, but I was not able to find anything that wasn't ```levelHigherOrEqualThan``` that seems a bit of the longish side... :confounded:

Spuds:
maybe minLevel and maxLevel ... not sure those make sense either

emanuele45:
Could make sense, I'll give it a try!

BTW: do **not** merge this branch until #2673 is verified and resolved, I would avoid changing the cache two times when it's broken.

live627:
Caught in testing hell? Ima fix that up for ya...

https://www.youtube.com/watch?v=k9Ylu0YywoA

emanuele45:
Something very like that. LOL

Spuds:
Looks like some shock therapy https://www.youtube.com/watch?v=nG2DwL3MCT4

emanuele45:
Such a long time I didn't see this episode!! LOL

emanuele45:
> maybe minLevel and maxLevel ... not sure those make sense either 

Let me see if this makes sense:
```
function minLevel($val) {...}
```
I would expect the function returns true if the current level is **at least** $val:
```
$cache->setLevel(2);
$cache->minLevel(1); // <= true
$cache->minLevel(3); // <= false
```

And the opposite:
```
function maxLevel($val) {...}
```
should give:
```
$cache->setLevel(2);
$cache->maxLevel(1); // <= false
$cache->minLevel(3); // <= true
```

What do you think?

Spuds:
seems more readable to me.

emanuele45:
I'm spending a lot of time trying to understand what I should put (if one or the other) in each case.
And if I'm having difficulties it means others will have as well... :-\

What would you think of something that reminds version_compare?
```
function levelCompare($val, $operator = '')
{
	if (empty($operator))
	{
		return version_compare($this->level, (int) $val);
	}
	else
	{
		return version_compare($this->level, (int) $val, $operator);
	}
}
```

Spuds:
That would work as well ... do we even need anything other than min level anyway?

emanuele45:
In theory we shouldn't need it, but I'm confused right now... lol
But... if we need just one, I would then use a ```levelLowerThan``` and make it a strict comparison. I feel it's the most readable and is not **that** long.

sources/SiteCombiner.class.php
```diff
@@ -188,8 +198,18 @@ public function site_js_combine($files, $do_defered)
 	public function site_css_combine($files)
 	{
 		// No files or missing dir then we are done
-		if (empty($files) || !file_exists($this->_archive_dir))
+		if (empty($files) || !file_exists($this->_archive_dir) || !is_writable($this->_archive_dir))
```
Spuds: Looks like a simple new private _checkSpares($files) method would cut out the duplication 😇

emanuele45: Yeah... look at these two methods makes me :sob:

install/Install_Controller.php
```diff
@@ -783,10 +783,6 @@ private function action_databasePopulation()
 		}
 		$replaces['{$default_reserved_names}'] = strtr($replaces['{$default_reserved_names}'], array('\\\\n' => '\\n'));
 
-		// If the UTF-8 setting was enabled, add it to the table definitions.
-		if (!empty($databases[$db_type]['utf8_support']))
-			$replaces[') ENGINE=MyISAM;'] = ') ENGINE=MyISAM DEFAULT CHARSET=utf8 COLLATE=utf8_general_ci;';
```
live627: The default charset may not always be utf8...

sources/SiteCombiner.class.php
Spuds: They did get the uglier over time, almost to the point of being difficult to follow 😢 I just did not anticipate all the complications we would need to account for.

install/Install_Controller.php
emanuele45: True, but the same answer I gave at https://github.com/emanuele45/Dialogo/commit/bb16cdc4e74435a5f44aced379d0e169a568edfc#commitcomment-19058260 applies.

sources/SiteCombiner.class.php
emanuele45: Who could have? lol

A bit of redesign could improve things, but I guess for the time being a method could be enough.

install/Install_Controller.php
live627: so the code got duplicated?

emanuele45: No, it is not replaced at all.
In the past the queries were written to end in:
```
) ENGINE=MyISAM;
````
but in the new version the queries end in:
```
) ENGINE=MyISAM DEFAULT CHARSET=utf8 COLLATE=utf8_general_ci
_______________^
```
see the missing semicolon that makes the replace useless?

live627: omg I didn't notice that before

Issue #2673, opened 2016-09-06
Problem with cache?


emanuele45:
I feel there is something not working properly with the cache (at least level 2).
* file-based cache
* log out from the forum
* manually delete all the json files
* log in
* log out
* log in
=> expected: be logged in
=> observed: still guest

If the cache is deleted it's possible again to log in.

Similar situation:
* file-based cache
* log out from the forum
* manually delete all the json files
* log in
* log out
* wait a few seconds (60?)
* log in
=> expected: be logged in
=> observed: a series of undefined index errors apparently related to groups not being loaded properly

PR #2674, opened 2016-09-07
Don't assume cached files are present

live627:
#2673

### DIFF
--- a/sources/subs/Cache.class.php
+++ b/sources/subs/Cache.class.php
@@ -415,6 +415,10 @@ class Cache
 	{
 		global $boardurl;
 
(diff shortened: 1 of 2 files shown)


sources/subs/Cache.class.php
```diff
@@ -415,6 +415,10 @@ protected function _build_prefix()
 	{
 		global $boardurl;
 
+		if (!file_exists(CACHEDIR . '/index.php'))
+		{
+			touch(CACHEDIR . '/index.php');
```
emanuele45: Reminder for later (because at the moment it doesn't work), instead of touch'ing it, it would be nicer to use ```secureDirectory``` that re-creates it "properly" along with the htaccess.

PR #2676, opened 2016-09-12
Inline perms

live627:
Closes #2663


sources/subs/InlinePermissionsForm.class.php
```diff
@@ -0,0 +1,245 @@
...
+	 * @param string[] $permissions
+	 */
+	public function save()
+	{
+		global $context;
+
+		$insertRows = array();
+		foreach ($this->permissionList as $permission)
+		{
+			if (!isset($_POST[$permission]))
```
emanuele45: We have an HttpReq object hanging around that is usually preferred to $_POST. :angel:

sources/subs/InlinePermissionsForm.class.php
```diff
@@ -0,0 +1,245 @@
...
+
+		loadLanguage('ManagePermissions');
+		loadTemplate('ManagePermissions');
+
+		// No permissions? Not a great deal to do here.
+		if (!allowedTo('manage_permissions'))
+			return;
+
+		// Load the permission settings for guests
+		foreach ($this->permissions as $permission)
```
emanuele45: Sorry for being a P.I.T.A.
At some point, it was kind of agreed to start using brackets around any loop/condition, even if one-liner.
In these cases with somehow long statements, in particular, it would be nice to start using them. ;)

emanuele45:
54 is indeed an odd commit message. :stuck_out_tongue: 

If I got it right, the admin-mp token is no more used, right?
Then, would it make sense to remove the bits from the templates (Admin and ManagePermissions) as well?

emanuele45:
BTW: **very** cool stuff!! :grinning:

sources/subs/InlinePermissionsForm.class.php
live627: ok then...

live627: 7afdabd1d4114c2a8846e5d5566f2644b8e386b7

live627:
> 54 is indeed an odd commit message. :stuck_out_tongue: 

27

emanuele45:
John Rayes wrote on 13/09/16 06:31:
>
>     54 is indeed an odd commit message. 😛
>
> 27
>
>

When 42? :P

emanuele45:
And the token in ManagePermissions? :angel:

live627:
still there

emanuele45:
Is it completed now?

live627:
Yes

Issue #2620, opened 2016-06-28
1.1.0 beta1: Personal message rules do not allow >1 criteria or action


tomreyn:
While defining a personal message filter rule, the 'Add criteria' and 'Add action' buttons do nothing. It is thus not possible to have more than a single criteria or action in a filter rule, although the UI suggests otherwise. 

This is on Firefox 47.0  with NoScript and RequestPolicy add-ons (as well as a few others).

On Chromium 50, in private mode (have not tested standard mode), I can have multiple criteria but not multiple actions.

tomreyn:
Note that I used the following value in some field to test susceptibility to code injections - this _may_ have triggered a bug on the web application making it behave erratically: 
```
'`"-->aaa<bbb>ccc
```

emanuele45:
The buttons normally work only with javascript enabled, seeing you have noscript, it may be you have js disabled.

Anyway, "add criteria" button works fine here, while the "add action" doesn't work even with js enabled.

tomreyn:
I had noticed and had whitelisted my Elkarte test install. I shall re-test this once there is another pre-release to test against.

live627:
This issue covers two bugs, actually; a JSON conflict and a cache trap. I have no idea how to fix the cache thing.

live627:
I noticed @emanuele45 fixed the cache issue in 2b505f876a0ab2a31ca5ed669c641f5c03e7eb53 but didn't merge it in.

live627:
And I tried with `'`"-->aaa<bbb>ccc` and it works fine.

emanuele45:
Double ff didn't close the issue. :P

PR #2680, opened 2016-09-14
JSON doesn't want to parse objects twice

live627:
#2620


live627:
I cherrypicked 2b505f876a0ab2a31ca5ed669c641f5c03e7eb53 because I really want to clean the issues list asap.

PR #2681, opened 2016-09-14
user offsets can be floats

### LINKED ISSUE #2597: Wrong value type sent to the database. Integer expected. (time_offset)
Elkarte version 1.0.7 / 1.1

There is a bug in function UserStatsPostingTime in Stats.subs.php
There is a line like: 
`'time_offset' => (($user_info['time_offset'] + $modSettings['time_offset']) * 3600),`
and query is expecting result to be integer
`{int:time_offset}`
but time_offset can be saved in feature settings as float value  (I'm using -0.004)
so If we visit user stats: /?action=profile;area=statistics;u={userId}
we have beautifully styled critical error:
`Wrong value type sent to the database. Integer expected. (time_offset)`
Fix is trivial to typecast it in query to (int)

PS. I remember fixing it in SMF the same way

live627:
fix #2597

### DIFF
--- a/sources/subs/Stats.subs.php
+++ b/sources/subs/Stats.subs.php
@@ -908,7 +908,7 @@ function UserStatsPostingTime($memID)
 	// Find the times when the users posts
 	$result = $db->query('user_activity_by_time', '
 		SELECT
-			HOUR(FROM_UNIXTIME(poster_time + {int:time_offset})) AS hour,
+			HOUR(FROM_UNIXTIME(poster_time + {float:time_offset})) AS hour,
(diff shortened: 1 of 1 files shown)


emanuele45:
@Spuds was that one to fix or to leave? Remembering the discussion I'm not sure at that point...

Spuds:
We either had to fix to be consistent between the form and the db ...  both either floats or ints instead of the mix we now have.   Floats I suppose allow more flexibility, even though the usage case seems a bit edgy, so its the way to go.

PR #2686, opened 2016-09-18
More duplicates

live627:
I found these by accident

### DIFF
--- a/sources/Errors.php
+++ b/sources/Errors.php
@@ -29,6 +29,8 @@
  * @param string|boolean $error_type = 'general'
  * @param string|null $file = null
  * @param int|null $line = null
+ * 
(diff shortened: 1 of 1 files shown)


Spuds:
I think we can blame @emanuele45 for that ... when the static class was added, the old error.php file was removed, but is now back https://github.com/elkarte/Elkarte/commit/25c545afab3d6e7fb7ebe1b895b0559bc3d0d336

In retrospect, having this old file done as you have is the proper thing to do for 1.0 compatibility, but it should be marked as depreciated in 1.1

live627:
deprecated and rebased

emanuele45:
I guess it's the usual thing when merging the 1.0 branch... :sob:

PR #2689, opened 2016-09-19
Update/break up Settings_Form

live627:
For feedback

Things to do:
- [x] Modify controllers to use Settings_Form::prepare() and Settings_Form::save()
- [x] Deprecate old API (the static methods)
- [x] Write tests (must always)

Things to consider:
- How useful are the hooks in Settings_Form? I  removed them...
- Add methods like setUrl() and setTitle() and setHook() for the controller to call?


sources/subs/SettingsFormAdapter.class.php
```diff
@@ -0,0 +1,85 @@
...
+	 * @param array $post_vars
+	 */
+	public function setPostVars($post_vars)
+	{
+		$this->post_vars = $post_vars;
+	}
+
+	public function __construct()
+	{
+		$this->post_vars = $_POST;
```
emanuele45:
```
$this->post_vars = HttpReq::instance();
```

sources/subs/SettingsFormAdapterFile.class.php
```diff
@@ -0,0 +1,491 @@
...
+	 * @var array
+	 */
+	private $new_settings = array();
+
+	/**
+	 * Helper method, it sets up the context for the settings which will be saved
+	 * to the settings.php file
+	 *
+	 * What it does:
+	 * - The basic usage of the six numbered key fields are
```
emanuele45: While you are at it, could it be worth creating 6 named constants like it has been done in other places?
Of course, in most of the cases we will still use the "natural" php filling, but having them could be helpful, even only for documentation purposes.

Something like:
```php
const VARNAME = 0;
const LABEL = 1;
const SAVETO = 2;
const DATATYPE = 3;
const DATASIZE = 4;
const HELP = 5;
```

emanuele45:
I guess I can set it for B3.

Spuds:
Awesome stuff 👍
Looks like its haveing trouble with the new testcase
```
1) TestSettingsForm::testInit
Undefined index: time_offset
/var/www/tests/sources/subs/SettingsForm.Test.php:45
```

emanuele45:
:open_mouth: 

Another thing to consider (it can be done after) is to move all the classes to a ```SettingsForm``` subdirectory and use a namespace like ```ElkArte\SettingsForm``` or ```ElkArte\sources\subs\SettingsForm``` (a bit verbose, but more consistent with the others... or maybe we can change the others xD)

live627:
>  Looks like its haveing trouble with the new testcase

Yep.... and I rewrote those tests last night. (still need to push)

live627:
> use a namespace 

Hey, good idea!

sources/subs/SettingsFormAdapter.class.php
live627: Uses $_POST because the controller may have modified it, and HttpReq does not contain the newly modified information.

Spuds:
I don't think that global search and replace spacing was a good thing to do ...  or that it should be part of this PR TBH.  It also has a very good change to cause merge issues in other open PRs.

I'm fine with function (bla) or funciton(bla) ... we can update as we go, just like with {}'s around everything.  Change them or not when you have that code open.

live627:
I'm done coding this, just need to do some final testing.

Spuds:
function () vs function() conflicts due to the SVG PR being merged in.  Rebase and resolve will clear those 3 up.

live627:
done

emanuele45:
> I'm done coding this, just need to do some final testing.

Then I'll wait until you give the green light. :wink:

live627:
yes, go ahead and merge this. We can always fix things later should there be an issue.

PR #2691, opened 2016-09-20
Some Template Cleaning / FA conversion to IcoMoon

Spuds:
This moves the SVG's from font awesome ones to IcoMoon which are "crisper" .. Some FA ones still remain as they were a bit better than the IcoMoon ones, but is 95% changed.  The ones that remained I did update the path definitions using the next higher level precision,  this makes them sharper but of course a bit longer path definition.

I also figured out how to properly define a gradient fill on the CSS SVG definitions, so for light the board icons are more like 1.0 in color

Several of the other small icons were green or blue, I changed these to be generic grey to prevent color overload, to much green or blue is not a good thing.

Changed over a few more of the source PNG callouts to CSS classes, more to go at some point.

BeSocial was lagging on 1.1's improved color / structure separation, so this helps to catch it up a bit.


Vekseid:
I didn't use gradients because I wanted to limit confusion. Best used sparingly until people acclimate. >_>

The reason I used the Font Awesome set was the lack of an attribution requirement. Since Icomoon combines sets, and we are distributing this, you will want to attribute the appropriate icon (and make sure the license is alright for us to distribute said icon) in a comment alongside it.

Spuds:
I only used icons from the icoMoon Free set ... it appears to be dual licensed GPL / CC BY 4.0, so. I think all I have to do is add them to the credits page like we have FA today.  Could add a comment to the CSS to indicate which is from which as well.  FA fonts are SIL OFL 1.1 licensed.

Yup on the gradients, I only used it in one place and thats the board icons for light.

Vekseid:
Cool cool.

Putting the icons used in the comments is helpful, I think. So far anything that I didn't have a comment for (that you haven't changed) was the icon of the same name from the Font Awesome set, and those I did comment referenced Font Awesome's label.

Spuds:
I've now remove all ```<i class="fa .....``` and moved them to our icon class

There are just a few Font Awesome calls remaining in the CSS as a font face .... I'll remove those next.   My question then becomes, since we are free of using FA in the core, should we stop loading it by default and instead maybe just load it via a modSettings that an addon can activate if needed but depreciate it?

Vekseid:
There were also some places that loaded it in script... I think I killed all of those, but double-checking may be good.

emanuele45:
Is that ready to go?

Spuds:
There is still more to go, but not for this PR

PR #2696, opened 2016-09-21
Admin help for modules is not called

Spuds:
I saw the methods in the modules themselves, but could not find how they were being called.  I added some code to call them, but I'm shooting a bit in the dark LOL ... @emanuele45 take a look and see if this makes sense or is even needed !


live627:
Looks good to me, although I haven't tested it.

I'd much rather see the regular triggers used.

Spuds:
*nods* ... triggers is likely the way to go, I'll close this one and open a bug report so it can be fixed proper like.

PR #2698, opened 2016-09-23
Convert Menu.subs to a class structure, for consideration

Spuds:
This one was done in the most primitive way, bit of a straight refactor to a class, nothing fancy.   It all started because of ```if ($admin_include_data == false)``` so I went to see if it was really always a bool being returned, the rest is a quick rebuild since it was ugly.  Frankly so is the class but its the start of a path we can continue in 2.0.


sources/subs/MenuCreate.class.php
```diff
@@ -0,0 +1,727 @@
...
+ * license:    BSD, See included LICENSE.TXT for terms and conditions.
+ *
+ * @version 1.1 beta 2
+ *
+ */
+
+/**
+ * This class implements a standard way of creating menus
+ */
+Class Menu_Create
```
live627: Why not just "menu"?

Spuds: I'd have to rename the file as well for the autoloader to pick it up.  But really I had no reason for the name other than that was the primary function name.  Be glad to rename it.

emanuele45:
>  It all started because of ```if ($admin_include_data == false)```

/me knows the pain! LOL

At the moment I just skimmed through the diff, so I'm entirely sure my comment will make sense, but since it may not make sense even if I read carefully the code, I'll write it anyway. :stuck_out_tongue: 

Would be a pain arrive to something like:
```
$menu = new Create_Menu($menuOptions);
$menu->addAreas($admin_areas);
...
$menu->addAreas($some_other_areas);
$menu->removeArea($area_id);
$menu_context = $menu->prepareMenu();
$menu->destroy();
```
the idea being: make the handling of menu areas separated from both the preparation and the instantiation of the menu.
What I'm after is being able to extend the menu at the "object" level instead of having to rely on the array. At this moment it doesn't really give many advantages, but in the middle/long term it should give some benefit (e.g. not having to pass-by-ref the whole array but instead pass the reference to the object).

live627:
I really like that idea to make the menus more modular, similar to what I did in https://github.com/live627/ModHelper/blob/master/src/Menu.php

Also @Spuds can you add tests for the menu?

Spuds:
> Also @Spuds can you add tests for the menu?

Not yet 😇 can do ... I did at least update some of the doc comments at the top as they were missing several of the available options.

> Would be a pain arrive to something like:

Of course it would 😛 I did say it was a primitive break up, you know when you find a 300 long line 10 deep nested if statement type of thing.  I'll add a couple of other methods, but I'm not going to go to crazy on this one at this time.

sources/subs/MenuCreate.class.php
Spuds: rename done !

Spuds:
While consolidating the permission check ( https://github.com/elkarte/Elkarte/pull/2698/commits/4a95203b14fb176b7384acbb2bc825d667a79b52 ) I think there was an old error in the code https://github.com/elkarte/Elkarte/blob/development/sources/subs/Menu.subs.php#L119 Looking at that line it calls $area['permissions'] but area is not set ... I feel it should be $section['permissions'] at that point.  I tried to track it down and it seems to have appeared when the first attempt at moving this 🐷 to a class was reverted.

sources/subs/Menu.class.php
```diff
@@ -161,9 +246,11 @@ private function _validateData()
 	private function _integrationHook()
 	{
 		// Allow extend *any* menu with a single hook
-		if (!empty($this->_menuOptions['hook']))
+		if (!empty($this->menuOptions['hook']))
 		{
-			call_integration_hook('integrate_' . $this->_menuOptions['hook'] . '_areas', array(&$this->_menuData, &$this->_menuOptions));
+			global $modSettings;
+			$modSettings['integrate_admin_areas'] = 'blabla';
```
live627: ?

sources/subs/Menu.class.php
```diff
@@ -766,4 +853,17 @@ public function callMenu($selectedMenu)
 			$selectedMenu['function']();
 		}
 	}
+
+	/**
+	 * Retrieve easily the sole instance of this class.
+	 *
+	 * @return Menu
+	 */
+	public static function instance()
```
live627: Singleton? Really?

Spuds: Yup ... best I can do!

I'm going to finish what I have with some test cases and then close this PR, I really don't think its worth changing all this ATM.

Spuds:
LOL .. failed tests, runs on my local though, sometimes Travis is a PITA

Anyway should someone want a go at this I'll leave the branch in my repo, but I'm closing this one out.

Note that I did add a few public methods addOptions and addAreas  which allow you to add / replace data in the menuOptions and menuData arrays.  

I get that everyone wants more flexibility in this class, but I don't have the passion, all I really wanted to due was break it down to components this round.

emanuele45:
Spuds wrote on 25/09/16 19:38:
>
> LOL .. failed tests, runs on my local though, sometimes Travis is a PITA
>

Yup.
I'm not sure why from time to time it happens.. :-\

> Anyway should someone want a go at this I'll leave the branch in my
> repo, but I'm closing this one out.
>

Is there any reason not to merge it?

> I get that everyone wants more flexibility in this class, but I don't
> have the passion, all I really wanted to due was break it down to
> components this round.
I'm more than happy with what you did!! It's wonderful to see code
become understandable! :D
Mine was an idea like I throw every time.
And any time I propose something is never with the intention of forcing
anyone to do anything (also because most of the times my ideas are at
least awkward). If I really feel it's something I'd like to see, I'll do
what I propose myself (unless someone beats me).

Spuds:
Your ideas were good, and I partially implemented them.

Then looking at things, in order to make it truly flexible, with it really needed a ground up redo vs the refactor to a class that I did.  Doing a ground up is not something I'm interested in for a little used function 😇 

So I figured why change it now when it would get replaced anyway, its not a step forward then, just one sideways.

Anyway the work is in my repo, the tests, meh on the tests, no idea.  It worked fine locally, but failed on Travis.  Then I tried a couple of things to no avail and probably broke it to boot!

live627:
Test failed because PHP died. Apparently, Travis doesn't like PHP 4 style constructors.

Spuds:
Thanks!  A little renaming in the test fixed it.

live627:
np

emanuele45:
BTW:
http://xkcd.com/1739/
:joy_cat:

Spuds:
LOL ... so many WTF moments

emanuele45:
And the hover is the best! lol

PR #2701, opened 2016-09-26
A continuation of #2696

### LINKED ISSUE #2700: Module help is not loaded into admin search
There methods in the modules themselves, but they are not being called.  They should load / unload with module activation. https://github.com/elkarte/Elkarte/pull/2696 for extra info

live627:
Closes #2700

I also went ahead and moved the module menu code.

...just did this real quick, not tested.


emanuele45:
The path I've been following is to use hooks if they are present, though it's not the best either because hooks have to be added/removed each time...
I guess as long as it works one way or another could do. :P

PR #2710, opened 2016-10-03
Some more fixes

emanuele45:


Spuds:
feel free to grab this as I'm lazy ATM https://github.com/Spuds/Elkarte/commit/6e8014eab6147bae39b6f5e83034c4a910cfbf9e 

When toggling between wizzy and normal, lists were getting a type=none at least with Vivaldi .. seems .css was returning 'none' when list-style-typet was not set/  Of course none is a valid value, which we interpret as remove the bullets.  Had to rely on using .attr first to check it it was really set, then .css to get the value.

live627:
I ran into that issue about 4 times, but kept forgetting to report it...

PR #2715, opened 2016-10-07
Some more fixes for B3

emanuele45:


emanuele45:
I changed this PR from a bug-fix to a feature-enhancement introducing a dynamic "max year" for the calendar. Tired of having to update cal_maxyear every now and then, I reached the conclusion that fix the calendar limit to 10 years from date('Y') is less hassles. :P

PR #2720, opened 2016-10-11
Implemented strict boolean comparison

Newman101:
Hello.

Looks like there were some non-strict boolean comparisons. In order to mitigate bug risks and improve performance, I implemented strict comparison to booleans. 

Thank you.

### DIFF
--- a/SSI.php
+++ b/SSI.php
@@ -186,7 +186,7 @@ function ssi_recentPosts($num_recent = 8, $exclude_boards = null, $include_board
 	{
 		$include_boards = is_array($include_boards) ? $include_boards : array($include_boards);
 	}
-	elseif ($include_boards != null)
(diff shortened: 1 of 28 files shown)


emanuele45:
I thought I created an issue for that but apparently I didn't. Oh well.

Thanks!
I reviewed quickly the changes, it seems fine. @Spuds any doubt? (I'm asking just because sometimes this kind of comparisons may have some edge case, like the failing test above.)

Spuds:
I took a look and they all appear safe to change, (null, true, false) comps ... 👍 Thanks @Newman101 Newman101

Newman101:
You're welcome. There are couple of edge cases because strict comparison (identity) returns true if variables being compared are equal and they are of the same type. I can have a look at these later but so far these implemented comparisons are safe to merge.

PR #2724, opened 2016-10-13
Errors in log

emanuele45:


emanuele45:
@live627 I had to work a bit on the cache class because of a fatal error in the admin panel 8a066dd7714534cc6a291aa8aced1cbdac97ad5d
I wonder if ```isEnabled``` should be used in some cases instead of ```isAvailable```. Dunno.

PR #2725, opened 2016-10-14
* I think the calls to topicPointer() have been wrong for a while

live627:
🍒 from #2333


emanuele45:
It's already in from #2641 ;)

live627:
Oh, sorry, I missed that.

PR #2682, opened 2016-09-15
compatibility with JQuery3

live627:
Replaced deprecated $.(un)load, $.(un)bind with $.on and $.off

PR #2726, opened 2016-10-14
Update editor again

Spuds:
We have been running a post version of SCE 1.4.7 in 1.1, basically the master branch with one commit reverted that was causing unnecessary white space removal.

An official 1.5.0 has been released, it should be good with jquery 3 (save that for b4) and fix some issues with MS edge, better source startup mode, etc.  Several (ok most) of the bugs reported on the master branch were not fixed, as such I had to add a hack to keep white space in code block intact when toggling source<->wizzy.

Add the cache stale to the wizzy style sheet, I thought this had been done, and updated the wizzy font stack to match the changes in the core font stack.


live627:
#2682

PR #2732, opened 2016-10-21
Some fixes

emanuele45:


emanuele45:
Apparently the PHP developers decided to change the behaviour of filter_var on booleans starting from version 5.6.27 (thank you for changing the behaviour of a function that was "broken" since 5.4 on a micro version).
I took the opportunity to move the version checks from the tests to the data validation class, forcing it to behave the same irrespective of the PHP version use (I hope). That, to me, seems a bit more logical behaviour than let the users of the class deal with the pain of php versioning.

PR #2733, opened 2016-10-23
A button to clear the CSS/JS hives (sort of maintenance task) + ElkInfoBar a bit of javascript to avoid re-creating self-hiding bars at the top of the screen

emanuele45:


emanuele45:
ElkInfoBar could be a sort of duplicate of the modal dialog, of the ajax notice, etc., but it does something slightly different from those, so... well I was bored.
Anyway, consolidate is the role of 2.0. :stuck_out_tongue:

Spuds:
Looks nice !

emanuele45:
This should do since there is the same structure just few lines above. :innocent:

emanuele45:
BTW: did I already say how much I like travis and automated testing in general? :smile_cat:

live627:
Was that sarcasm? I honestly can't tell...

sources/SiteCombiner.class.php
```diff
@@ -262,6 +262,42 @@ public function getSpares()
...
+
+	/**
+	 * Deletes hives from the cache based on extension.
+	 */
+	protected function _removeHives($ext)
+	{
+		global $context;
+
+		$path = $this->_archive_dir . '/hive-*.' . $ext;
+		$names = array();
```
live627: unussed

emanuele45:
> Was that sarcasm? I honestly can't tell...
>

No, it wasn't! :sob:
That one just now saved me the pain of a patch after the release! (I
wonder if anyone is going to test use on 5.3 anyway, but that's another
problem lol.)

PR #2737, opened 2016-10-27
Fix search engine url

interlab:
Now when you add URL addresses in will not be replaced by the symbol **_&_** for the new search engine.


emanuele45:
Thanks!

PR #2742, opened 2016-11-14
Split a query for SSI recent topics

live627:
Split that fugly query yet again. This is some code I made for a client a while ago. Originally for SMF, but I  imagine that Elk uses the same process here. **This is untested.**

This query could be further refined, the joins on the members table could be replaced with `loadMemberContext()`, though that may not be ideal as it can add too much complexity, not sure if worth it.

### DIFF
--- a/SSI.php
+++ b/SSI.php
@@ -432,27 +432,46 @@ function ssi_recentTopics($num_recent = 8, $exclude_boards = null, $include_boar
 	if (empty($topics))
 		return array();
 
+	$topic_list = array_keys($topics);
+
+	// Count number of new posts per topic.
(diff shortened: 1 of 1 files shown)


emanuele45:
> This query could be further refined, the joins on the members table could be replaced with ```loadMemberContext()```, though that may not be ideal as it can add too much complexity, not sure if worth it.

Yeah, loadMemberContext is always a bit borderline when it comes to optimization.

About the PR, seems good, let's wait a couple of days to see if Spuds has any comment. :angel: 

BTW, would you mind using two different names for the two $request? For example the first one could be $topic_request. I think I tried since a while to avoid using the same name within the same function.

Issue #2741, opened 2016-11-11
1.1B3 Forum maintance long working time


gevv:
Hi,

Forum maintance "Find and repair any errors"  takes too long 1 hour more    (It was a minute in v1.0.9)

Vekseid:
Could you tell what step it was stuck on?

gevv:
- Topics with incorrect first or last message entries
- Topics with the wrong number of replies
- Topics with the wrong unapproved post count
- Messages linked to non-existent members
- Topic logs linked to non-existent topics 
- Topic logs linked to non-existent members

and after

Vekseid:
As in it hung for an unreasonable amount of time for each of those?

emanuele45:
Try this, in RepairBoards.subs.php, change:
```
	if (!$force && microtime(true) - $time_start > 3)
```
to:
```

emanuele45:
Fixed in https://github.com/elkarte/Elkarte/pull/2751/commits/3243421254bbb6a554a3904af5170324ad610ca1

PR #2751, opened 2016-11-22
Some more or less random things as we clean up for an RC

Spuds:
@live627 I had to make a couple of minor adjustments due to changes with the new settings class.  

One was the birthday email previews were not displaying as var_message was removed from the template, I changed that back https://github.com/elkarte/Elkarte/pull/2751/commits/c4cb2b6daea764ac59321113b5f251356e80a9db

Second was the cache description was not displaying, for that I had to use 'desc' twice in the config vars array https://github.com/elkarte/Elkarte/pull/2751/commits/e31e031d0657897c71bb127156e9d653ab29aae3 Not sure those are the correct things so would appreciate if you could take a look and see if something else should be done instead of what I did.


emanuele45:
@Spuds would you mind grabbing the fix at #2741 ? :angel:

Spuds:
Got it, will update the PR soon

Spuds:
Nuff for now I think

Spuds:
And now I'll fix the conflicts, happy days

PR #2754, opened 2016-11-23
 many random unused things that Scrutiniser found 

live627:


sources/controllers/Auth.controller.php
```diff
@@ -29,11 +29,7 @@ class Auth_Controller extends Action_Controller
 	 */
 	public function needSecurity($action = '')
 	{
-		if ($action === 'action_keepalive')
-		{
-			return false;
-		}
-		return true;
+		return !($action === 'action_keepalive');
```
emanuele45:
```
return $action !== 'action_keepalive';
```
Isn't it easier to read? :angel:

emanuele45:
That github projects thingy just above labels and milestones is a bit of a pain with the rest of the UI elements jumping up and down while you are using them... very badly designed. **nods*

sources/controllers/Auth.controller.php
live627: Well, why didn't I think of that before!

Another PR incoming soon...

PR #2760, opened 2016-11-25
Parsemail

Spuds:
Found a problem with a specific email which had a multi-part inside of an multi-part (thanks outlook, you crazy sob)

The outer was multi part related and inside were 3 sections, one of which was a multi part encoding.  The class was parsing correctly but due to the nesting the inner multi was being overwritten by the outer (blank) text.  This should fix that and if we are lucky have no ill effects 🙏 

Couple of other minor tweaks since I had the file open, like lots of {} love.


live627:
Selenium strikes again!

Spuds:
Be nice if it at least gave a reach around :p  ... Restarted

sources/subs/PbeImap.class.php
```diff
@@ -91,14 +91,7 @@ public function __construct()
...
 		// I suppose that without this information we can't do anything.
-		if (empty($this->_hostname) || empty($this->_username) || empty($this->_password))
-		{
-			return false;
-		}
-		else
-		{
-			return $this;
-		}
+		return (empty($this->_hostname) || empty($this->_username) || empty($this->_password)) ? false : $this;
```
live627: Constructors do not have meaningful return values, anything that is returned from here is discarded

https://scrutinizer-ci.com/g/elkarte/Elkarte/inspections/cb1de7e7-f5f8-49d2-9fad-7efcff3e78ab/issues/files/sources/subs/PbeImap.class.php?status=new&orderField=path&order=asc&honorSelectedPaths=0

Spuds: Guess thats why it was so fugly before !

Spuds:
Don't merge this just yet ... found some more buglets in this function.

PR #2764, opened 2016-11-28
A few errors fixed, plus a couple of tests

Spuds:


tests/sources/ProfileInfoTest.php
```diff
@@ -0,0 +1,81 @@
...
+		global $context, $modSettings;
+
+		$controller = new ProfileInfo_Controller();
+		$controller->pre_dispatch();
+		$controller->action_index();
+
+		// Lets see some items loaded into context, there should some data
+		$this->assertNotNull($context);
+		$this->assertEquals($context['can_see_ip'], true);
+		$this->assertEquals($modSettings['jquery_include_ui'], true);
```
live627: you can use assertTrue() here.(same for other files)

tests/sources/subs/BBCHTML.subs.Test.php
```diff
@@ -0,0 +1,64 @@
...
+
+	/**
+	 * Prepare what is necessary to use in these tests.
+	 *
+	 * setUp() is run automatically by the testing framework before each test method.
+	 */
+	public function setUp()
+	{
+		$GLOBALS['modSettings']['user_access_mentions'] = array();
+		$GLOBALS['modSettings']['enablePostHTML'] = 1;
```
live627: Surely this (using `$GLOBALS`) can be avoided...

tests/sources/ProfileInfoTest.php
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	/**
+	 * Test the settings for admin search
+	 */
+	public function testProfileI()
+	{
+		global $context, $modSettings;
+
+		$controller = new ProfileInfo_Controller();
```
live627: Seems to me that we need a controller factory object to aid in the creation of controllers.

PR #2765, opened 2016-11-29
Search methods updeate

live627:
Simplifies code a little and make the API namespaces PSR-4 compatible.

Existing API calls should be returning the same results.


emanuele45:
Conflicts...
Please try to use rebase and not merge. ;D

emanuele45:
BTW, after that, I'd say let's keep the refactorng for 2.0, so we can flush out 1.1 ASAP. ;D
If we are confident, we could even branch 1.1 at the first RC and keep active development out of the release game, and at each step we can re-merge the fixes from 1.1 into dev the same way we do for 1.0 patches.

live627:
Rebase completed.

live627:
> If we are confident, we could even branch 1.1 at the first RC and keep active development out of the release game, and at each step we can re-merge the fixes from 1.1 into dev the same way we do for 1.0 patches.

I like that idea!

PR #2767, opened 2016-12-01
Some fixes, some formatting, some tests

Spuds:
Few more poorly written tests just to exercise the code, we do need some improved structure
A few minor "bug" fixes,
Minor refactor of the paypal controller since it was a pending todo (broke out a couple of methods from one of the jumbo ones we have), 
Couple of bugs as reported in .12
Some things for scruit
Some formatting, love the {}'s


Spuds:
Yeah, I broke selenium again!

Spuds:
At least mysql 5.7 is running now, must have been broken for some time as it was falling back to 5.6.  Selenium is extra mad now that I pointed it to chrome, if anyone has some ideas other than stop messing around with selenium :P

emanuele45:
Could it really be that Chrome is not installed?
I think it is not shipped with ubuntu by default.

I found:
http://blog.500tech.com/setting-up-travis-ci-to-run-tests-on-latest-google-chrome-version/
and follwed a bit, though Trusty seems to break apache https://travis-ci.org/emanuele45/Dialogo/builds/180552900 :confounded:

Spuds:
Interesting .. I finally did get chrome running based on the above, but it requires an ass load of deps to install as well, so just a PITA

I originally went to chrome as FX was giving some "random" failures.  However updating FX as part of the selenium based testcase install and using the latest chromedriver inside of the head appears to have fixed the issues.

I'll update the PR soon.

Spuds:
Yeah it passed baby .. the sql versions are correct as well !

PR #2769, opened 2016-12-02
Fix and test internal admin search

live627:
Some labels weren't resolved

### DIFF
--- a/sources/admin/Admin.controller.php
+++ b/sources/admin/Admin.controller.php
@@ -65,8 +65,25 @@ class Admin_Controller extends Action_Controller
 		$context['robot_no_index'] = true;
 
 		// Need these to do much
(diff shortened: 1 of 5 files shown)


live627:
Huh... the tests are running but no coverage increase? Wha?

Spuds:
Some coverage when down ... 

I think since the coverage test (which is the php 5.6 run) did have some errors (selenium again), perhaps that caused the coverage report to miss a few items.   

I restarted it and it ran fine.  It may be that the selenium tests need a short delay added to the browser and head to give it a chance to respond before items are checked?

emanuele45:
I forgot about this one... can be merged?

live627:
ofc

live627:
@emanuele45 you up late xD

PR #2770, opened 2016-12-03
Grinding away!

Spuds:
Added a implicitWait to web tests in hopes it stabilizes it.
Change quote logic/css 
Convert a few more icons to svg
ILA was not using the mime image in posts only preview
tweaked the attachment headers a bit,  empty Pragma was just strange and Content-Transfer-Encoding: binary for only gecko must be old or just cargo coding.


emanuele45:
Fine to bring it in?

Spuds:
I think so, easy enough to open another one ... right now I'm trying to get a v1.5.1+ version of the editor working, that will be another PR anyway.

emanuele45:
I wasn't sure if you wanted to test anything on the wait for the tests. :D

Spuds:
So far adding the implicit wait seems to have helped with the selenium fails 🙏

PR #2773, opened 2016-12-06
Update editor to "latest"

Spuds:
We had 1.4.7 with some, of what eventually became 1.5.0, fixes pulled in

Here is 1.5.2 with some of what (appears) will be v2.0 cherry picked.  It does not include the flexbox pr to the editor since that seems to break the layout for us and I don't want to deal with it ATM. 

I did pull in some of the other 2.0 fixes (like the code whitespace issue and ie8 support removal) otherwise 1.5.x is a no go for us.  Anyway God help us all.


live627:
/me quietly looks around

#2726

/me slips into the shadows

emanuele45:
1.5.2 > 1.5.0 ... is it not?

Spuds:
This is 1.5.2 along with some of what will be in 2.0 pulled in (such as the official fix for the code white space stripping).  I see Sam has pushed a couple of more "2.0" fixes that I'll also pull back in.  Should official 2.0 escape soon enough then we can consider using it.

The current version in the repo is a 1.5.0-ish with the previous fix I had used to prevent the white space removal.  

Prior to that it was a version of 1.4.7 (in the 1.1 line) with most of what ended up in 1.5.0 except for the JQ3 fixes I think, so really it was a non-official 1.5.

I'm sure thats not any more clear, but there you have it.

Spuds:
Also do we want to change over to jQuery 3 as well?

emanuele45:
Spuds wrote on 07/12/16 15:45:
>
> Also do we want to change over to jQuery 3 as well?
>
>

I thought we did that already... That gives you the general idea of how
I feel about javascript. LOL

If it will not break (a lot of) random stuff sure! (If it breaks just a
little who cares :P)

Spuds:
I think we have 2.1.4 which I did some time way back in the development branch, 1.0 is still 1.x I think?? Anyway I'll give it a test today and see if there are any noticeable issues, I think the major work has been done to make it ready for 3.1, unless someone has already done that 🙏

PR #2787, opened 2016-12-14
Patching a bit the protocol stuff

### LINKED ISSUE #2783: Links inserted into posts with http/https in uppercase
as reported here http://www.elkarte.net/community/index.php?topic=4149.0

links inserted with HTPP or Https are broken

emanuele45:
fixes #2783

### DIFF
--- a/sources/Subs.php
+++ b/sources/Subs.php
@@ -1793,9 +1793,16 @@ function isValidEmail($value)
  *                 added if none is found (optional, default array('http://', 'https://'))
  * @return string - The url with the protocol
  */
-function addProtocol($url, $protocols = array('http://', 'https://'))
+function addProtocol($url, $protocols = array())
 {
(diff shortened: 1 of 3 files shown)


sources/Subs.php
```diff
@@ -1793,11 +1793,26 @@ function isValidEmail($value)
  *                 added if none is found (optional, default array('http://', 'https://'))
  * @return string - The url with the protocol
  */
-function addProtocol($url, $protocols = array('http://', 'https://'))
+function addProtocol($url, $protocols = array())
 {
-	foreach ($protocols as $protocol)
+	if (empty($protocols))
+	{
+		$pattern = '~^(http://|https://)~i';
```
Spuds: Nothing to change, but just wanted to note how this vs https?:// works ... if you feed each with https://asfdasfasd whats there will take 15 steps to find the "hit" since it has to fail and then backtrace to the start.  The one I just typed, should only take 9 steps as it will not require a backtrace.   Either would take 9 steps if fed http.

Issue #2744, opened 2016-11-17
Fixes to backport from 1.1 to 1.0.10


emanuele45:
- [x] #2709
- [x] #2541

Optional:
- [x] #2600
- [x]  #2746

emanuele45:
#2746 to consider

PR #2796, opened 2016-12-28
Patch 10

### LINKED ISSUE #2766: Full editor does not work on ios 10
We need to package up a new version of the editor for 1.0.10 so that it works with newer iOS 10 on phone/tablets.  A fixed version of the editor is available at this post.

http://www.elkarte.net/community/index.php?topic=4006.msg29619#msg29619

Spuds:
Some updates for 1.0.10 ... 
Some Updates for #2744 did not wrap eval in catch statements for php7 and have not taken care of the image rotation update
Fix #2766 editor with ios (this is not the new editor as in 1.1, but a patched version of 1.0.9)
Other fixes as indicated


Spuds:
Something going on with SensioLabsInsight .....

emanuele45:
Yep, seems they are having some issue with our builds... let's see if in a few days it comes back to life, otherwise I'll ping them and see.

PR #2810, opened 2016-12-30
2017 Preparation Update

Trekkie101:
- Remove redundant calendar entries
+ Add St. Andrew’s Day
* Clarify which independence day
* Update copyrights


emanuele45:
/me takes note we should implement a better way to deal with this kind of stuff... :-\

Thanks!

Trekkie101:
For the copyright dates, it could just be date('Y') or remove them entirely, not sure they're really relevant but helps show how long the project has been about. Only flaw with date() is a misconfigured server. 

Potentially for the holidays:
1) We could over engineer a tool to make the recurring but not 'set' dates.
2) Remove them - since I doubt anyone uses them
3) Remove non 'set' dates and instead add more fixed date holidays.

cody3647:
Would there be a way to use relative date formats...that would cover most non-set date holidays I would think.

emanuele45:
When I wrote the comment I was thinking about a kind of flexible relative date based on month/weeks, what you are used to see in a calendar app: set an even every 1st Sunday of the month, every 1st Sunday of July, on Tuesday every other week, etc.

PR #2820, opened 2017-01-02
Fixes for 1.1 RC1

emanuele45:


themes/default/Post.template.php
```diff
@@ -476,6 +476,7 @@ function template_add_new_attachments()
 		numOfAttachmentAllowed: ' . $context['attachments']['num_allowed'] . ',
 		totalAttachSizeUploaded: ' . (isset($context['attachments']['total_size']) && !empty($context['attachments']['total_size']) ? $context['attachments']['total_size'] : 0) . ',
 		numAttachUploaded: ' . (isset($context['attachments']['quantity']) && !empty($context['attachments']['quantity']) ? $context['attachments']['quantity'] : 0) . ',
+		fileDisplayTemplate: \'<div class="statusbar"><div class="info"></div><div class="progressBar"><div></div></div><div class="control fa fa-times-circle"></div></div>\',
```
Spuds: can we use one of the icon class ones (SVG) instead of fa ?

emanuele45: Very well I guess.
This was just a copy&paste from the js file of code written a while ago.
Any hint since I don't remember the code to use? lol

Spuds: Should be something like
```class="control icon i-close"``` (not sure what the control class is), but the other are basically fa => icon and fa-times-cirlce => i-close

Comment (emanuele45) [inline themes/default/Post.template.php]: Done (mostly) https://github.com/elkarte/Elkarte/pull/2820/commits/f15862286ad201c211c95a32a8bdb6556ac03c8f
There is only the http://fontawesome.io/icon/share-alt-square/ left, not sure what I can use in its place.

**Comment by emanuele45** [comment]
Also, is this https://github.com/elkarte/Elkarte/pull/2820/commits/185b6c8998b0717a59def97e2f75869d5bb966d9 okay?

**Comment by Spuds** [comment]
For the share icon
```
.i-share::before {
	content: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' stroke='%23EEEEEE' stroke-width='.07em' fill='white' viewBox='0 0 32 32'%3E%3Cpath d='M27 22c-1.4 0-2.7.6-3.6 1.53L9.95 16.8c.04-.27.06-.53.06-.8s0-.53-.02-.8L23.4 8.5c.92.94 2.2 1.52 3.6 1.52 2.76 0 5-2.24 5-5s-2.24-5-5-5-5 2.24-5 5c0 .27.02.54.06.8L8.6 12.52C7.67 11.6 6.4 11 5 11c-2.76 0-5 2.24-5 5s2.24 5 5 5c1.4 0 2.7-.58 3.6-1.52l13.46 6.73c-.04.3-.06.57-.06.8 0 2.8 2.24 5 5 5s5-2.2 5-5-2.24-5-5-5z'/%3E%3C/svg%3E");
}
```
for both light and besocial css ... we will probably need to adjust the .abort, .remove, .share font size and padding so these ILA icons size and pad correctly.  Yes the  ::before was missing, good catch !

emanuele45:
Spuds wrote on 06/01/17 02:16:
>
> For the share icon
>

Done, thanks!!

> for both light and besocial css ... we will probably need to adjust
> the .abort, .remove, .share font size and padding so these ILA icons
> size and pad correctly.

Guess so, but... ehm... maybe it's better if someone other than me does
this. lol

> Yes the ::before was missing, good catch !

That was firefox that ignored the image because of the missing ::before.
O:-)

PR #2822, opened 2017-01-04
fix bug profile fields changes

### LINKED ISSUE #2593: change gender
ElkArte 1.1

index.php?action=profile;area=forumprofile

you have a sex change in the profile (maybe only my problem)

interlab:
fix #2593

### DIFF
--- a/install/install_1-1.php
+++ b/install/install_1-1.php
@@ -2384,7 +2384,7 @@ class InstallInstructions_install_1_1
 				array('warning_watch', '10'),
 				array('warning_moderate', '35'),
 				array('warning_mute', '60'),
-				array('admin_features', ''),
+				array('admin_features', 'cp'),
 				array('last_mod_report_action', '0'),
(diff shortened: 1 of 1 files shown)


emanuele45:
The main issue was that they were set enabled in the install script f1521b6f22fc253b9f86f9403b8afecbb4c8da16

That said, I feel you are probably right saying the profile fields should be enabled by default.
Now I'm not sure if merge this one and fix it by enabling the fields and disable them... lol
Probably the best option is:
1) merge this and have CP on by default,
2) deactivate some of the existing fields (like most of the social sites and the very old AOL, etc.)

interlab:
do as you like. 
the gender of the user and the location can be done enabled by default.
don't know.

emanuele45:
Yes, that's what I thought as well.

emanuele45:
In 6a8c573bf099c10386ea0b4978b0a7d867913851 I disabled the "social" custom fields (kept active location, gender and alike).

PR #2823, opened 2017-01-06
Upgrade updates

Spuds:
Found a few minor things in the 1.1 upgrade ... most of that is 

https://github.com/Spuds/Elkarte/commit/431067464526d02185763a026b0d2eccbe5a9ffa the rest of this  PR is predominantly formatting except for one case where we appear to have overwritten $db

Now the odd thing is this runs fine from the command line but sometimes seems to "stall and die" with the web interface.  No idea why, its just part of the "wonder" that can be found in this script.

### DIFF
--- a/install/CommonCode.php
+++ b/install/CommonCode.php
@@ -139,7 +139,7 @@ class Ftp_Connection
 	}
 
 	/**
-		* Changes a files atrributes (chmod)
+		* Changes a files attributes (chmod)
 		*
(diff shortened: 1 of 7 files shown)


emanuele45:
In the DbTable class we have a table_exists method! :D 
So an:
```
if ($db_table->table_exists('{db_prefix}admin_info_files'))
{
   ...
```

should work... unless there is a but. :angel:

PR #2835, opened 2017-01-22
minor 1.0.10 adjustments

Spuds:
Couple of more items for 1.0.10 ... lets see if scruit fires on this one (which through much pain its working again)


emanuele45:
nope... :sob:

emanuele45:
I'm wondering if it makes sense to keep sensiolab when it fails every two or three checks... :-\

emanuele45:
BTW
/me feels it's time to start preparing the patch... :sob:

Spuds:
Seems sensiolab  only dies on the "master" or "patch" branches ... seems to run fine on development.  Yeah the patch should be getting close, there are the "other items" to add as well 🔒

PR #2845, opened 2017-01-29
git repos attempt to store the file cache for themes, fix?

FictionCircle:


FictionCircle:
Up to you, I can just fix it whenever I mess with it.

emanuele45:
It's right to add the exception, just could you use... dunno:
```
hive*
```
just to allow the index.php talready there?

cody3647:
The main cache directory isn't ignored either

FictionCircle:
Yeah I realized I forgot to copy that over. Oops. Should be fixed for both now.

emanuele45:
The main cache directory is ignored:
https://github.com/elkarte/Elkarte/blob/development/.gitignore#L4

cody3647:
Hmm.  Must have gotten confused there.

FictionCircle:
> cache/data*.php

https://github.com/elkarte/Elkarte/blob/8214f42dc6c370bfe60018b2517776b7b982798e/sources/subs/CacheMethod/Filebased.php#L46

https://github.com/elkarte/Elkarte/commit/85baa540d059416f0dedaf8a916b57d1c2af1296

> cache/data*.json

They were switched to .json files so its no longer ignored.

emanuele45:
Actually, I didn't notice the json switch. :-\
It's not really a good thing TBH.
The cache file area/were PHP as an additional line of defence against leaking of sensitive data (yes, the file names are random, and all, but misconfiguration may expose the directory listing and json files are plain text files that can be read from the net.

I'll do a quick restore for RC1, it may generate broken php files, but it doesn't really matter that much.

FictionCircle:
@emanuele45 

Yeah, I thought it was odd but I have .json setup to execute as .php so I wasn't worried about it. You could just leave it alone and just change the extension to .php instead of a restore (which I'm assuming means revert).

emanuele45:
Change the extension alone wouldn't make much difference, without a php opening tag. ;)
I'm adding to #2842

PR #2846, opened 2017-01-29
Add a #new line to separate posts

Spuds:
Not sure we want this, but it was a question on the forum.  I like it,  but then all my taste is in my mouth !

newnewline


cody3647:
Can't you just put the new id on the hr tag and get rid of the a tag.  Same for the message a tags as well.  Just put the id on the article tag.

Spuds:
you sure could

PR #2864, opened 2017-02-08
Change profile link to use href from memberContext

cody3647:
The href value is already there in the array and it makes it easier for a mod to change where the name and avatar link to


emanuele45:
I don't remember if there was a reason for that or if it was just overlooked, anyway it looks good, so merged. :P

PR #2865, opened 2017-02-09
Revert "file_exists(): open_basedir error fix"

Frenzie:
Reverts elkarte/Elkarte#2855

I don't believe this change in logic relates to the open_basedir error, but more importantly, it makes it so that add-ons don't show.

See http://www.elkarte.net/community/index.php?topic=4274.msg30708#msg30708


Frenzie:
Don't merge this yet if you were planning to. :-)

emanuele45:
Applied the fix at #2870, this can be closed.

Frenzie:
Good thing I forgot to add it this afternoon when I finally had some time! :-P

PR #2867, opened 2017-02-10
Return True Not False

ahrasis:
We have discussed this in here:
http://www.elkarte.net/community/index.php?topic=4274.0

My reasoning is based on the old code that says continue and that works
to show all the packages inside its directory.

I don't change the others i.e. regarding the extension since this is
already working fine (no error so far) so that can be considered not
broken.

I do maintain my earlier suggestion regarding the bracket position as PR earlier and as
explained in the said topic as well.

signed-off Ahmad Rasyid Ismail "ahrasis" ahrasis@gmail.com


emanuele45:
Applied the fix at #2870, this can be closed.

PR #2876, opened 2017-03-01
Holiday ids are submitted via post not get

cody3647:


Spuds:
Thanks, good catch.   We (OK I) messed a few of these up when converting over to using the httpreq class, especially when the old was the ambiguous REQUEST .

PR #2879, opened 2017-03-07
Parse ADDONSDIR in package info

live627:


emanuele45:
Apparently he doesn't know either. :P

Issue #2893, opened 2017-03-24
A non well formed numeric value encountered


ahrasis:
Detected while using php7.1 http://www.elkarte.net/community/index.php?topic=3952.msg31448#msg31448

Suggested solution by emanuele is here: http://www.elkarte.net/community/index.php?topic=3952.msg31496#msg31496

Spuds:
When https://github.com/elkarte/Elkarte/commit/1018fc8853bab752b4c375d83fd32af219f2ad62 is merged in from 1.0.10 will be the final parts of the fix

PR #2894, opened 2017-03-24
moar pain

Spuds:
Update cssmin to latest to fix php7 issue ... since this is an external the file can just be replace
Update script,js so jump box works on ios and is non blocking

### DIFF
--- a/sources/ext/cssmin.php
+++ b/sources/ext/cssmin.php
@@ -1,7 +1,7 @@
 <?php
 
 /*!
- * cssmin.php v2.4.8-4
+ * cssmin.php v2.4.8-5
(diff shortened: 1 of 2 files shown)


emanuele45:
Does that fix #2893 as well?
I'm not sure why we pass a path to CSSmin when it wants a boolean... :-\

Spuds:
It fixes the error he was getting, as php7. x is a bit tighter on some loose values.

But no it does not address the specific issue of passing  ```$this->_cache``` to cssmin.  But that also looks harmless as what is passed gets the ```(bool)``` treatment, so it will end up being true and that's the default.  So although thats an error (and should be fixed), it seems harmless, can wait till 1.1 unless we want to also toss it in 1.0 as well.

PR #2907, opened 2017-04-17
Corrected PHP Error when no custom profile settings are selected

tinoest:


emanuele45:
Thanks, sorry for the waiting but it got off the radar... :blush:

PR #2908, opened 2017-04-20
! bad backport from 1.1, or maybe bad in 1.1 as well :innocent:

Spuds:
Well pretty much screwed the 🐩 on that one ... Few other than me use this, but it should at least work.

@emanuele45 these two quick fixes should, if possible, be added to the official patch.  If its to late thats kind of fine as well, but I messed up bringing it in from 1.1, or I messed it up in 1.1 as well LOL


emanuele45:
Oook!
I have already updated the version number of the template on the branch I'm using locally, so let's keep it like that.

PR #2934, opened 2017-05-12
Added $db_type to bootstrap.php 

tinoest:
Fixed bug where its not set when calling bootstrap.php and therefor tries to use mysql as a default

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -31,7 +31,7 @@ const CACHE_STALE = '?R11RC1';
 // We're going to want a few globals... these are all set later.
 global $time_start, $maintenance, $msubject, $mmessage, $mbname, $language;
 global $boardurl, $webmaster_email, $cookiename;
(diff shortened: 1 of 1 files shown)


emanuele45:
Why that one was overlooked for so long?
Sorry @tinoest my fault... :blush:

emanuele45:
A suggestion: usually it's better to use rebase to bring a branch in line with the main development one instead of using merge, this because it makes the history much cleaner (less "merge *** branch" in the history).

tinoest:
Will try that next time, I'm still used to the svn way of merging and syncing with upstream if I'm honest.

emanuele45:
No problem. :)

If you want to compare it to the "SVN-way", I think it's just a matter of replacing update (?) with rebase. (Well, maybe not **exactly**, but something like that, with some jumping between branches in addition. lol)

PR #2951, opened 2017-06-24
More minor rc1 cleanup

Spuds:
Fix a few minor be social issues
Fix search regression where the "did you mean" link did not search for spellchecked term
Spacing fixes from previous doc block update
Misc


Spuds:
Travis is failing as it did in the other recent commit (looks like phpunit fails to start) *shrugs*

Spuds:
Doing a quick look, at some point we dropped the composer lock file ? ?  So it could be that a version of phpunit is being installed that is not compatible with 5.6, thats just a guess though.

emanuele45:
It seems it just fails because the php 5.6 package has problems:
https://s3.amazonaws.com/travis-php-archives/binaries/ubuntu/12.04/x86_64/php-5.6.tar.bz2
```xml
<Error>
  <Code>AccessDenied</Code>
  <Message>Access Denied</Message>
  <RequestId>086AD544CE81EABF</RequestId>
  <HostId>PZ4bm7DjoAc3EN8zgbid2Ghd8v0PzSw7RhCgJ/zBU2ELNUl/jiBURV1QKbfddkkV5QyOchXgB68=</HostId>
</Error>
```

Spuds:
That is new as of this AM .... it was failing last night with the 255 after phpunit is called, I tried again this AM and now you see the 5.6 failure.

I'm looking at whats required to update the entire thing to trusty 14.04 since the old environment is not going to be around forever.

emanuele45:
Any objection at merging even if travis is broken?

Spuds:
I think its fine to merge (as is yours)  ... I'm still working on getting the tests to run again, but the new trusty environment is causing me some other difficulties.  Also I'm going to drop 5.3 testing, will add 5.4 so we have some depreciated version but 5.3 is not available easily on the new environment, at least not that I've found

PR #2952, opened 2017-06-25
Some more bugz :P

emanuele45:
I wouldn't text the Imagick part... if you can please do so. :innocent: 
Also, while writing it, I had the funny idea to have an exception class able to render the error message as image. Maybe in the future.


emanuele45:
So, now that **is** funny:
Pull request builds fine https://travis-ci.org/elkarte/Elkarte/builds/247272491?utm_source=github_status&utm_medium=notification
My repo fails https://travis-ci.org/emanuele45/Dialogo/builds/247272441
Same code, just two builds few seconds apart. :tada:

PR #2954, opened 2017-06-29
Update test bed

Spuds:
This change updates the test environment from Precise (12.04) to Trusty (14.04) which is where everything will be moving.

This caused several changes and I took the "opportunity" to try and update some dependencies.

  - Tests on php 5.3 have been replaced with 5.4.  Not ideal to our min requirements but reasonably good enough to check if we are using anything to modern *cough*
  - PHP inside of Apache has been moved to use PHP-FPM processes as the old way was not supported on the new platform, at least as far as I could tell.  Bit of a PITA but what can you do
  - MySQL tests on 5.5 were dropped, only 5.6 and 5.7 are tested (5.5 may work with some effort but when I tested it, downgrading the server environment to 5.5 did not go well).  postgre is tested against 9.3, 9.4 and 9.5
  - PHP 5.6+ tests were moved to phpunit 5.x and selenimum 3.x.  PHP 5.4 still runs on phpunit 4.x and selenium 2.x as more recent versions of php unit don't support old versions of php.
  - Updated the webtests to use a newer (not the newest) versions of geckodriver and selenium, this was required as the old versions failed.  The new version however did not support ->keys("") so some tests had to be tweaked.
 - dropped PHP 7.0 and added PHP 7.1
  - Added a few more environment vars (coverage and webtests) to try an optimize the runs, there is no benefit (IMO) to running the web tests against more than one of the matrix tests, long runs and takes time to build up the server dependency list.  Non coverage runs (meaning it does not output % coverage reports to scruitinizer) turn off xdebug so they run much faster.

Thoughts?


Spuds:
And of course this ran fine against my fork ( https://travis-ci.org/Spuds/Elkarte/builds/248438601 ) but fails against the base repo.   Fuck you Travis ! 🖕 Yer a piece of shit 💩

Spuds:
Well it fails doing ```apt-get update``` ... must be something going so I'll try again later.

Spuds:
Appears Travis fixed the issue on their end sometime last night and the tests are now running again.

emanuele45:
So it's good to merge?

Spuds:
Should be .... do note we do loose php 5.3 and mysql 5.5 testing but *shrugs* I just don't see that as a big deal for a variety of reasons.

emanuele45:
Do we want to branch out 1.1 in order to keep at least some testing on them and have this in the development (2.0 to be) branch only?
I don't have very strong interests in 5.3 apart from the "it should work", or we can just drop support before releasing 1.1 on the basis that we do not have any test on it (i.e. saying: it **should** work, but if you find something broken do not blame us). :innocent:

Spuds:
The reality is the current tests cover less than 15% of our code base .... so even having 5.3 in the mix, what does it really indicate?  I think we hvae been caught more by compatibility issues with newer releases of php and sql then worrying about whats behind us 👀

emanuele45:
/me was trying to be picky :stuck_out_tongue:

Spuds:
If we find this causes problems, it should be *simple* to move it back to 12.04 and add back 5.3 .... simple within the confines of debugging Travis issues.   Anyway 1.1 is done in terms of big code changes 🙏 and we know rc1+ ran on 5.3 so its all good LOL

PR #2956, opened 2017-07-05
Fix some, add some

Spuds:
Few minor fixes and one minor "add" ... I'd detail them here for future generations to read, but why when you can just read the self explanatory, and creative narratives in the commits ?


emanuele45:
Just for the record (because I read it a couple of days ago), some wbmp related functions are going to be stripped from PHP 7.2.

Spuds:
Had to look up WTF that is LOL ... 

I'm not sure we use any of the specific conversion functions that are going to be removed, but looking at what it is (B/W BMP image), I'd say its safe to remove it from 1.1, can't imagine its being used at all.    Anyway really something for 2.0, the graphics support / util stuff needs to be in its own class.

Issue #2375, opened 2016-02-08
Replacing Font Awesome and Image Sprites


Vekseid:
https://github.com/Vekseid/FontReaper

After I get finished with the .css segregation, I was thinking of going through and excising every last sprite and other raster image from the css. Font Awesome has most everything as-is.

Given the issues icon fonts have, I wrote a small tool to make them into something a bit more robust. : )

emanuele45:
Sorry, these days I'm either totally :sleepy: or doing something else.
You mean replace sprites with FA or replace everything with something else?

Spuds:
I think the idea is to replace them with FA icons, but to use SVG versions of those icons ?

Vekseid:
Right. I made the script above, which dismantles and icon font and makes an embedded .css version that works across all modern browsers, and returns to the font fallback for IE 7-8.

Right now the situation is a bit ridiculous. You are loading Font Awesome, which has icons for nearly everything you use images for, and yet they go unused. >_>

Spuds:
Yup somewhat selective use to date ... and old discussion here: https://github.com/elkarte/Elkarte/issues/1321

emanuele45:
One question that came to my mind: you are converting to svg only part of the fontawesome stuff, correct?
So if someone want to use other icons the FA it may be worth considering something along the lines of the jQuery-UI stuff? (i.e. keep FA in the core and have a flag to enable it, mainly for addons I think.)

Vekseid:
Might be worth it to leave as a toggle, yeah. One of the perks of picking a new naming scheme means it is an option. : )

emanuele45:
Just to avoid opening another issue (unless you prefer so, in that case I'll do ;D), the unread and unreadreplies icons are black, on BeSocial becomes rather difficult to see them.
For a total newbie, how do I change the background color of the SVG?
Is it necessary to edit the path, or is it possible to style it somehow?

Frenzie:
You can style SVG through regular CSS. Actually I'd propose to add in relevant style hooks for easy styling through custom CSS without having to perform any JS hacks.

Specifically, you'd add a little stylesheet link in every SVG
```
<?xml-stylesheet type="text/css" href="whatever.css" ?>
```

Then it's probably best to work with some `class` attributes in the SVGs themselves.

Finally you'd have your `whatever.css` with a couple of quick styles like

```
.whatever {stroke:black; fill:white;}
```

That way you could easily adjust the SVGs to your theme.

[...]

PR #2960, opened 2017-07-09
svg the message index

Spuds:
This is part of #2375

This uses the SVG icons for the standard sprites topic icons used on the message index.  It does not change them elsewhere as they are part of menu picks or select pulldowns.  So a minor inconsistent between the message index and the message display.   

This changes the following for the message index: poll, clip, question, exclamation, last post, lamp, thumbsup, thumbsdown, lock, sticky, sort down/up ... basically most (not all) icons that were used from the topics/topicicons.png on the message index.  See the pic

download


Spuds:
I should really see if the rest of the topic icons in that sprite can be updated, I think there are only a couple left.  Then all the topic_icon class definitions could be removed from the css.   

This does keep the same css class names in the template, those could be changed to the common icon i-
 class name but was not sure if that would break custom themes.  :/

emanuele45:
I'm pretty sure we already broke quite a bit of themes anyway. lol

Spuds:
I'd do some more breaking then 😇

Spuds:
Think I now converted all of the topic_icons.png file to use the existing svg images (well i did have to add a couple).

custom_variant.css was loading before the variant.css preventing variant specific overwriting.

Moved the SVG css that was in the variant files to its own file and added the ability to have a variant version to allow overwriting.  This cuts down on the redundant css, but really it was more to allow a single file edit vs two files which was easy to forget.  Added a new function to load the svg, and custom css files etc.

Spuds:
No need to check for the file existence since loadCSSFile will do that for us.

custom.css would fallback to the the default theme ... I changed that since I think that file should be theme dependent and not falllback to any existing one in default or should it ??? Thoughts about that?

emanuele45:
Spuds wrote on 14/07/17 13:45:
>
> No need to check for the file existence since loadCSSFile will do that
> for us.
>
> custom.css would fallback to the the default theme ... I changed that
> since I think that file should be theme dependent and not falllback to
> any existing one in default or should it ??? Thoughts about that?
>

/me agrees

PR #2965, opened 2017-07-28
Add a select box to load the agreement in different languages while r…

### LINKED ISSUE #2892: Checkbox agreement should allow for language selection

emanuele45:
…egistering - fixes #2892

### DIFF
--- a/sources/controllers/Register.controller.php
+++ b/sources/controllers/Register.controller.php
@@ -83,6 +83,7 @@ class Register_Controller extends Action_Controller
 			'contact' => array($this, 'action_contact'),
 			'verificationcode' => array($this, 'action_verificationcode'),
 			'coppa' => array($this, 'action_coppa'),
+			'agrelang' => array($this, 'action_agrelang'),
 		);
 
 		// Setup the action handler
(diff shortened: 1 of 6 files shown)


emanuele45:
Okay, I'm a little lost here. :laughing:

Spuds:
You mean the Travis 5.6 error ?   Something changed when they began the Trusty rollout.  I had it all working just fine a week before this on the Trusty environment.  I'm frankly in no mood to debug that ATM, getting it to work at all is a dependency version level nightmare.

emanuele45:
I tried changing some versions, but I didn't get anything.
I'll revert the changes and give up for the moment.

Spuds:
Believe me I understand the frustration with getting it to run.   It took quite a bit of effort to get the combination that WAS working with trusty .. the number of incompatibles with different drivers and browsers and php unit selenium is crazy stupid.

themes/default/scripts/register.js
```diff
@@ -414,4 +414,30 @@ function onCheckChange()
...
+					$('#agreement_box').html(request);
+				}
+			})
+			.fail(function(request) {
+				// Maybe reload with language attribute?
+			})
+			.always(function() {
+				// turn off the indicator
+				ajax_indicator(false);
+			})
```
Frenzie: You forgot the semicolon here. :-P

themes/default/scripts/register.js
```diff
@@ -414,4 +414,30 @@ function onCheckChange()
...
+	$(document).ready(function() {
+		$('#agreement_lang').change(function() {
+			$.ajax({
+				type: "POST",
+				url: elk_scripturl + "?action=register;xml;api=json;",
+				data: {sa: 'agrelang', lang: $(this).val()},
+				beforeSend: ajax_indicator(true)
+			})
+			.done(function(request) {
+				if (request != '')
```
Frenzie: Is that safe or should it be something like `if (request && request != '')`?

sources/controllers/Register.controller.php
```diff
@@ -1258,4 +1265,42 @@ public function action_registerCheckUsername()
 
 		$context['valid_username'] = !$errors->hasErrors();
 	}
+
+	public function action_agrelang()
```
Frenzie: Why not `action_agreelang`? Does that already exist? Something against double vowels? :-P

themes/default/scripts/register.js
emanuele45: It's a foreseen exception. :-P

themes/default/scripts/register.js
emanuele45: hmm... I don't **think** request can be not set and it should at least be a string (though jquery documentation is not exactly easy to read).

sources/controllers/Register.controller.php
emanuele45: Yeah, I was considering action_agreementlang but it seemed a bit on the longish side.
agreelang looks indeed slightly better than agrelang.

emanuele45: Actually, I realized there is a better place for localization-related stuff, so I moved it there. I had to move a bit of other stuff around, but I think it looks nicer now... maybe. :stuck_out_tongue:

PR #2971, opened 2017-08-12
Few fixes

Spuds:


emanuele45:
I'll ignore codacity because I don't understand the complain. lol

Spuds:
Yeah it probably wanted a typeof bla !== 'undefined'  for robustness ... we have a few of those that we could change ... not now though :P

PR #2974, opened 2017-08-18
Bug fix and some features that will help SimplePortal avoiding code edits

emanuele45:


sources/subs/Cache.class.php
```diff
@@ -57,6 +57,12 @@ class Cache
 	protected $_accelerator = null;
 
 	/**
+	 * Cached keys
+	 * @var string[]
+	 */
+	protected $_cached_keys = array();
```
albertlast: i would define this as static,
otherwise you could get issue when the same time this logic run twice.

Some addition checks in construct etc. would be than needed too.

sources/subs/Cache.class.php
```diff
@@ -417,6 +434,39 @@ public function remove($key)
...
+	 */
+	public function removeKeys($keys_match, $delimiter = '~', $modifiers = '')
+	{
+		if (!$this->isEnabled())
+		{
+			return;
+		}
+
+		$to_remove = array();
+		foreach ((array) $key as $k)
```
Spuds: $keys_match or $key ?   .... per codacy

emanuele45: :innocent:

sources/subs/Cache.class.php
albertlast: Nice idea with the reduce time overlapping.

emanuele45: The class is instantiated inside a singleton, so it should not run twice (unless some external code instantiate a new caching method by itself, in which case I'm not sure I would have this variable shared).

TBH, this is not the best solution either, I would have preferred if each caching had it's own way to deal with removing single keys, but it's not the case. From my readings it is reliably possible only for the file-based caching (because you can go through the file names), that is the less relevant, while the others have **some** way to remove **some** keys, but no way to find and remove keys based on pattern recognition. :disappointed:

PR #2975, opened 2017-08-22
A little experiment

emanuele45:
What do you think about it @Spuds ? :angel: 

```
$value = $AbstractModel->_modSettings->aProp;
```
$value either gets the value of aProp, or null

```
$value = $AbstractModel->_modSettings->aProp('default_value');
```
$value either gets the value of aProp, or 'default_value'


Spuds:
I like it!  It also makes it similar to how we interact with the request object for $_stuff

PR #2984, opened 2017-08-29
Prevent deletion of board when it still has posts or topics.

### LINKED ISSUE #2872: Posts in deleted boards
Just nixed this on Elliquiy - having it set so that a board can't be deleted if it has any topics.

Any thoughts about me prettying it up (making a setting, etc) and putting it into a PR? The changes are fairly unintrusive.

Vekseid:
Closes #2872


emanuele45:
Could you please move the query to a function and move the function to... for example Boards.subs.php?
Queries should not be present in controllers. :wink:

Vekseid:
Found a function I could use >_> Though I think the naming convention is a bit confusing.

emanuele45:
>  Though I think the naming convention is a bit confusing.

Strangely enough it was not my doing in case. :laughing:

PR #2988, opened 2017-09-09
Small language change

d3vcho:
"Powered by" and "Credits" looks very bad with their first letter in lowercase.


emanuele45:
WTF... @Spuds does that mean selenium is broken again?

Spuds:
Sure looks like that ... what a royal PITA !

emanuele45:
What a f****ing pain...
What I don't understand is why it breaks even though nothing really changed: the versions are the same as last week. :-\

Spuds:
Its friggin infuriating to say the least.

PR #2996, opened 2017-09-25
Other fixes

emanuele45:


emanuele45:
For the moment I can't see anything else critical for 1.1.
If it's like that, I would release before mid-October. What do you think @Spuds ?

Spuds:
We should include the update @Vekseid is working on so the accessibility items are cleaned up.  Then there are a few mobile things to tweak but nothing major.  We can target mid October to start of November time frame.  And as allways we need to "check" the error log for any goodies!

emanuele45:
> We should include the update @Vekseid is working on so the accessibility items are cleaned up.

Right, I was looking at the issue and forgot about it. :roll_eyes: 

> And as allways we need to "check" the error log for any goodies!

Doesn't apply the "if you didn't see it, it didn't happen"? :angel: :laughing:

Vekseid:
PR for basic fixes submitted : p

PR #2997, opened 2017-09-26
Assorted critical accessibility fixes

Vekseid:
Some cosmetic changes

- Aria-role menuitem was uselessly noisy. Instead made sure it got labelled in a way that mobile screenreaders can use.

- Return focus to verification box after the user selects 'listen to'. Got props for. 

- Swapped legacy object code out for the audio tag in non-IE browsers. This fixes some, but not all issues with iOS Safari. Further work is needed.

- Some basic attempts to help make likes and polls stand out a bit. May need a larger test group than my current trio.

There were a number of other attempts at fixes, but they did jack so are not included.


Frenzie:
Nice!

PR #3036, opened 2017-10-29
Allow webtests to fail

live627:
I want my green checkmark dammit!


emanuele45:
Spuds already disabled entirely the webtests in 1.1 at 594bd7cd59d60799601e7448d604efc1b14fc97c

PR #3038, opened 2017-10-29
Remove any compatibility hack with the 1.x line from 2.0

emanuele45:


install/install.php
```diff
@@ -81,9 +81,6 @@ function initialize_inputs()
 	if (ini_get('magic_quotes_sybase') || strtolower(ini_get('magic_quotes_sybase')) == 'on')
```
live627: kill this too

install/installcore.php
```diff
@@ -13,7 +13,7 @@
 const CURRENT_VERSION = '1.1';
 const CURRENT_LANG_VERSION = '1.1';
 const DB_SCRIPT_VERSION = '1-1';
-const REQUIRED_PHP_VERSION = '5.3.3';
+const REQUIRED_PHP_VERSION = '5.5.0';
```
Frenzie: If this is the case you should also update the README.

emanuele45: Good catch, done!

install/install.php
emanuele45: I think I left it as a warning, but... yeah, doesn't make much sense I guess.
Removed.

emanuele45:
@live627 I saw the SessionHandler class, if I understood correctly, with PHP 7.x the whole class can just be removed and used the PHP built in, right?

live627:
PHP 5.4, but yes, drop in replacement.

emanuele45:
Good, then I'll remove it as well before merging the branch I guess.

emanuele45:
I removed everything I could find along with all the hacks for any php < 7... I think. :laughing:

PR #3040, opened 2017-10-30
Restore some API that was lost during development

live627:


emanuele45:
I thought it was removed intentionally... oh well. :stuck_out_tongue: 
Just minor consideration from my side: )I would prefer to have the frontPageHook as "preferred" method of replacing the front page (because it's the new one), and the integrate_action_frontpage as (deprecated) new one. In other words, I would move the integrate_action_frontpage block before the frontPageHook if.

Another note: when you use a comment like "Compatibility" or "For old time's sake", please add also the version you want to be compatible with, otherwise in few years from now we will be there asking ourselves "which one was the version that needed this compatibility hack?". :wink: 
Something like:
```
// Compatibility with 1.0.x
// For old (1.0.x) time's sake
```
It would be super to mark both the beginning end the end of the compatibility code, but I guess it's seldom possible...

PR #3050, opened 2017-11-08
[wip] Refactor the menu

live627:
This gives the menu system a complete overhaul while remaining completely backwards compatible. Each menu item is an object to help ensure everything that we want is in a proper format.

Obviously still needs work. Tokens and session checks in the profile, for example, are not being set.

Request for comments.


Spuds:
Sorry for not having had time to look this over just yet :( .... still banging away on 1.1.1 

But on the bright side I've also been busy with a new truck and finally a new laptop to replace my 6 year old one.  Now given enough time I might be able to find all my software keys :P  

Anyway this is something we need to do.   I see you removed all the _ things, in general we have tried (not very consistently) to use the _ when its a private/protected method.  I'll try to pull this in to a test branch to provide feedback, just swamped ATM.

live627:
neat. I'm also busy so am taking a break from this

PR #3053, opened 2017-11-13
Use Bootstrap in index.php

Spuds:
There was *a lot* of duplicate code between index.php and bootstrap ... this is a *first* pass at using a single bootup class for both SSI and "normal" elk to remove the duplication.

This may be a bit much for 1,1.1 ... if so we need to still make a couple of minor changes to allow SSI to load from inside ELK ... so things like portals work 😸 

If we want to go this way i'll add in some quick docblocks, right now this was really a quick and dirty setup so things should work in 1.1.1 and we can do some proper things in 2.0.  Also if so PLEASE test this out, I have not checked the logs (site or server) for issues just yet. 

If however we feel this is to much, I need to make a commit to fix the SSI loading form inside ELK, so let me know which way we feel is the right way (for 1.1.1 that is)

### DIFF
--- a/SSI.php
+++ b/SSI.php
@@ -12,7 +12,7 @@
  * copyright:    2011 Simple Machines (http://www.simplemachines.org)
  * license:        BSD, See included LICENSE.TXT for terms and conditions.
  *
- * @version 1.1
+ * @version 1.1.1
  *
  */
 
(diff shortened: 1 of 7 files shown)


emanuele45:
Neat! :smile_cat:

live627:
Less duplicate code means less bugs.

emanuele45:
I'm a bit unsure about the last commit: 86e289111eb595090bea0bb47d1364f03274aff4
My original idea when I moved from SSI to bootstrap in these files was that bootstrap **is** the way to initialize the system, while SSI is a kind of demo showing how to use bootstrap and providing few custom functions to work with and since in these files the functions from SSI are not needed, they would just use bootstrap.

Spuds:
Fair enough, I actually had not tried any of those functions on just bootstrap.  

I reverted them back to require bootstrap and then added a  ```new Bootstrap()``` call.  

That call could also be added to the end of the bootstrap file, and if we do that probably should add in a big ass global blablabla statement as well so those are available when the require is done and not like what I did in subscriptions (which I think is necessary if I've got my head around the top level stuff)

Spuds:
Added a standalone param to the constructor, this should make ```new Bootstrap``` in theory just what it was/is in 1.0.  Calling it as ```new Bootstrap(false)``` should allow it to be used from index, loading just the basics that are common between normal and side entrance startups.  Calling SSI will call bootstrap assuming it has not been called before (allowing it to be included inside of ElkArte.  I suppose SSI could simply become an API to provide addon capability once ElkArte has been loaded.

PR #3078, opened 2017-12-16
 Remove more compatibility code

live627:


bootstrap.php
```diff
@@ -118,7 +118,7 @@ private function setRusage()
 	/**
 	 * If they glo, they need to be cleaned.
 	 */
-	private function clearGloballs()
+	private function clearGlobals()
```
Spuds: The doc block is less fun now :P

PR #3081, opened 2017-12-18
$message_labels needs to be iniitialized per pm

live627:
if the message has no label attahed, $message_labels was never set


emanuele45:
Could you please send the PR to the patch_1-1-2 branch?

live627:
Done!

PR #3082, opened 2017-12-18
Refactor/js

live627:


Spuds:
Early Christmas :D .... Travis error is a simple escape

```unexpected 'autosuggest_delete_item' (T_STRING) in /var/www/test/themes/default/ManageMembergroups.template.php on line 386```

PR #3083, opened 2017-12-18
Update version headers

live627:
Bettere late than never, eh? Also, this will clear up some confusion regarding which branch to use.


live627:
Rebased.

Spuds:
can't see this PR causing the problem, but the error is:
```
1) TestAuth_Controller::testLogout
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'An Error Has Occurred'
+''
```

live627:
Yes. Totally unrelated error.  I've fixed it locally and will send in a PR soon.

PR #3089, opened 2018-01-02
Refactor/sql

live627:
This is the first step in reworking some SQL queries to run  a bit faster (filesorts are bad, m'kay). These  are the queries I was able to get through these past few days. I  plan to do more in the future once I have some more time.

Several workarounds for pgsql work just fine in mysql, so I hardcoded them in.

### DIFF
--- a/sources/subs/MessageIndex.subs.php
+++ b/sources/subs/MessageIndex.subs.php
@@ -168,6 +168,7 @@ function topicsParticipation($id_member, $topic_ids)
 		FROM {db_prefix}messages
 		WHERE id_topic IN ({array_int:topic_list})
 			AND id_member = {int:current_member}
+		GROUP BY id_topic
 			LIMIT ' . count($topic_ids),
 		array(
 			'current_member' => $id_member,
--- a/sources/subs/MessageIndex.subs.php
+++ b/sources/subs/MessageIndex.subs.php
@@ -35,7 +35,6 @@ function messageIndexTopics($id_board, $id_member, $start, $items_per_page, $sor
 	$db = database();
 
 	$topics = array();
-	$topic_ids = array();
 	$indexOptions = array_merge(array(
 		'include_sticky' => true,
 		'fake_ascending' => false,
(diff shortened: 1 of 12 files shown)


live627:
Anyone still alive here?

Spuds:
Still alive, but in the funk !

albertlast:
I'm waiting for intresting stuff,
many of the stuff is already done in smf
Some other solution aren't better as the existing one.

Like this one: https://github.com/elkarte/Elkarte/pull/3089/commits/232dad99d713b45ca5b7056161ef716a098e2f84
yes both query are faster as one, but this are two queries which take in sum the same amount of time as one query,
that was the result in my test env.

emanuele45:
hmm... I didn't get any notification from github about this one... :-\

sources/subs/MessageIndex.subs.php
```diff
@@ -168,7 +168,6 @@ function topicsParticipation($id_member, $topic_ids)
 		FROM {db_prefix}messages
 		WHERE id_topic IN ({array_int:topic_list})
 			AND id_member = {int:current_member}
-		GROUP BY id_topic
```
emanuele45: Actually, I think you need it, or in alternative you need a DISTINCT because messages holds a lot of id_topic per each message and without the aggregation you potentially get an awful amount of results that will be looped both for fetching and for processing later on. :wink:

sources/subs/MessageIndex.subs.php
```diff
@@ -123,18 +111,11 @@ function messageIndexTopics($id_board, $id_member, $start, $items_per_page, $sor
 				LEFT JOIN {db_prefix}log_topics AS lt ON (lt.id_topic = t.id_topic AND lt.id_member = {int:current_member})
 				LEFT JOIN {db_prefix}log_mark_read AS lmr ON (lmr.id_board = {int:current_board} AND lmr.id_member = {int:current_member})') .
 				(!empty($indexOptions['custom_joins']) ? implode("\n\t\t\t\t", $indexOptions['custom_joins']) : '') . '
-			WHERE ' . ($ids_query ? 't.id_topic IN ({array_int:topic_list})' : 't.id_board = {int:current_board}') . (!$indexOptions['only_approved'] ? '' : '
-				AND (t.approved = {int:is_approved}' . ($id_member == 0 ? '' : ' OR t.id_member_started = {int:current_member}') . ')') . '
-			ORDER BY ' . ($ids_query ? 'FIND_IN_SET(t.id_topic, {string:find_set_topics})' : ($indexOptions['include_sticky'] ? 'is_sticky' . ($indexOptions['fake_ascending'] ? '' : ' DESC') . ', ' : '') . $sort_column . ($indexOptions['ascending'] ? '' : ' DESC')) . '
-			LIMIT ' . ($ids_query ? '' : '{int:start}, ') . '{int:maxindex}',
+			WHERE t.id_topic IN ({array_int:topic_list})',
```
emanuele45: Shouldn't this have an order (at least in PHP) somewhere?

emanuele45:
Since you are at it, I think 'alter_table' is not used anywhere. :angel:

sources/subs/MessageIndex.subs.php
live627: I'll reverse that change, then.

sources/subs/MessageIndex.subs.php
live627: The order is in the array, set by the first query.

live627:
> Like this one: 232dad9
yes both query are faster as one, but this are two queries which take in sum the same amount of time as one query,
that was the result in my test env.

Page two and above used two queries anyway... and the goal there was to stop sortng by find_in_set.

> ilike is very very slow in pg,
so this "improvment" 1b8580c should run slower.

That should now behave the same as in mysql.

sources/subs/MessageIndex.subs.php
albertlast: The order in IN operator as fare i know not guaratee that the data is sorted by this.

albertlast:
The common approach is to lower both side,
could help you also in mysql when a user do wrong stuff in mysql and use a cs charcode.

sources/subs/MessageIndex.subs.php
live627: Correct. I mention the PHP array though.

live627:
Could you explain what you mean by that please?

emanuele45:
@live627 the notifications coming from you were caught in the spam filter. :laughing:

albertlast:
`where col1 ilike '%aSdf%'` -> `where lower(col1) like lower('%aSdf%')`
in my pg benchmark this solution runs faster on pg as ilike and work on mysql and pg

emanuele45:
BTW I think the insert_into_log_messages_fulltext and the insert_log_search_results_subject "NOT\s" replacements are not happening anyway, because to me the query has two spaces between the "NOT" and the "LIKE"/"RLIKE" and the replacement only one.
Fortunately, the "pure" LIKE/RLIKE replacements are coming to the rescue.

Aaaanyway @live627 do you plan to tackle the conversion of ilike to lower as well or I can just merge?

live627:
Thanks @albertlast for the hint

emanuele45:
Thanks!

PR #3090, opened 2018-01-02
Support MS-LCID

live627:


sources/Subs.php
```diff
@@ -451,6 +451,11 @@ function standardTime($log_time, $show_today = true, $offset_type = false)
 
 	$str = !is_bool($show_today) ? $show_today : $user_info['time_format'];
 
+	// Windows requires a slightly different language code identifier (LCID).
+	// https://msdn.microsoft.com/en-us/library/cc233982.aspx
+	if ($is_win)
```
emanuele45: Just a little tweak: would you mind always using brackets even for oneliners? :innocent: 
```
if ()
{
    
}
```

emanuele45: Just a bit more context: let's say this is the direction used since a while, because it's more consistent and even though it uses a couple of lines more, it's in several cases much more readable than the other version in several cases.

live627: sure, done

emanuele45:
:sunglasses: 
Thanks!

PR #3094, opened 2018-01-15
Master

ahrasis:
https://www.elkarte.net/community/index.php?topic=4867.0

Slogan doesn't replace logo fix.

### DIFF
--- a/themes/default/index.template.php
+++ b/themes/default/index.template.php
@@ -238,7 +238,7 @@ function template_body_above()
 
 	echo '
 				<span id="logobox">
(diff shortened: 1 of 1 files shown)


emanuele45:
Actually, the thing to fix is the text of the string so that it doesn't say anything about replacements. :smile_cat: 
Also, please use the branch https://github.com/elkarte/Elkarte/tree/patch_1-1-2 as starting point and send the pull request there (I think it is worth a fix in 1.1.2), thanks!

live627:
can this now be closed?

PR #3101, opened 2018-01-21
Remove/unused

live627:
PhpStorm code inspection result


emanuele45:
You always upset codacity, you are a bad guy. **nods** :laughing: 

BTW, I'm working on another branch where I actually use $db_prefix passed to elk_db_initiate to avoid that global in the database class.... actually this is not entirely true, I used it **somewhere**, but not exactly in the initiate function... or at least I think I used it somewhere...
What do you think: global or parameter?

live627:
Probably a parameter so that it cans be assigned as a property for the
methods that use it.

PR #3105, opened 2018-01-25
Where are my icons? I NEED MY ICONS!!

live627:


emanuele45:
May that apply to 1.1 as well?

live627:
No, icons show there as evidenced by elkarte.net.

PR #3126, opened 2018-03-07
Add key to select and radio custom fields

emanuele45:
This PR tries to add the bases to custom profile fields to provide better support for CSS (i.e. changing the field value for select boxes or radio buttons doesn't require changes to CSS class names) first and internationalization later. Introducing a key/value association for select boxes and radio buttons fields.
It should preserve backward compatibility by falling back to the previous behaviour if json_decode fails.

And in the meantime, I sneaked in the implementing of Frontpage_Interface for the Recent_Controller. :innocent:


emanuele45:
I wonder if it would make sense to hide by default the key and show it with an interaction (e.g. click) to avoid confusion, or if it is fine to keep it visible and editable.
One possibility could be to have it in a disabled field and automatically take the value of "value" when creating a new entry, and provide an "edit" button for editing of existing values (both on adding new and editing).

Issue #3124, opened 2018-03-07
Undefined index: unread_messages


Spuds:
Notice: Undefined index: unread_messages
Undefined index: unread_messages
PHP Fatal error: Uncaught exception 'ErrorException' with message 'Undefined index: unread_messages' in /var/www/sources/controllers/PersonalMessage.controller.php:307
```
Stack trace:
#0 /var/www/sources/controllers/PersonalMessage.controller.php(307): error_handler(integer, string, string, integer, array)
#1 /var/www/sources/controllers/PersonalMessage.controller.php(224): _messageIndexBar(string)
#2 /var/www/sources/SiteDispatcher.class.php(359): action_index()
#3 /var/www/index.php(136): dispatch()
#4 /var/www/index.php(66): elk_main()
#5 {main}
thrown in /var/www/sources/controllers/PersonalMessage.controller.php on line 307
```

Spuds:
Appears to be due to ```$context['labels'] = loadPMLabels($context['labels']);``` which correctly has ```$context['labels'][-1]['stuff']``` set, but loadPMLabels does not use that passed value and instead starts off with an empty array ignoring what its passed.

emanuele45:
Ref:
https://github.com/elkarte/Elkarte/pull/3127#issuecomment-371308252
fa78e4b2db21ec6cdd7fe866a62839dfd831037f

PR #3127, opened 2018-03-07
Bring 1.1.2 changes to development

emanuele45:
I hope I didn't miss anything...


Spuds:
Its fa78e4b that causes #3124 ... it needs to be done in the function param ```$lables = array()``` instead else it overwrites what is passed.   I suppose technically it should be merged and then the 1.1.3 fix be brought in?

emanuele45:
I see... I didn't notice the rewriting. Darn.
Well, since we have to fix it in 1.1.3 I'd say so (it should be easier for the merging, even though it's not always so straightforward).

PR #3135, opened 2018-03-18
Fixes #3133 and update to BadBehaviour 2.2.20

### LINKED ISSUE #3133: BadBehaviour.php has backtick in SQL, this isn't valid in PostgreSQL
Line 212 and Line 356 have the backtick character in the SQL, this throws a error when using PostgreSQL. 

`$bb2_blocked = bb2_db_query('SELECT COUNT(*) FROM {db_prefix}log_badbehavior WHERE `valid` NOT LIKE \'00000000\'');`

I can either remove them from the SQL or check the DB Query for that character and remove it. 

Which would you prefer and I'll submit a PR for it.

tinoest:
Changes to update to 2.2.20 from 2.2.19 and removed backtick around SQL.  Fixes #3133 

DATE is a keyword in MySQL by it seems ok with the lowercase date on my version of MySQL. If there are issues I can add single quotes around instead.


emanuele45:
Is it worth having this in 1.1?

tinoest:
@emanuele45 was that aimed at me? 

It’s technically a bug fix rather than a enhancement, the updates were also to fix bugs reported in the upstream software.

tinoest:
I’ll switch to the other branch, do you want this PR left to sort the dev as well as the patch?

emanuele45:
Thanks!
I'll close this one when the other is merged (since the fix will come in development when the patch will be merged here).

PR #3137, opened 2018-03-18
Removed backticks and updated to latest version of badbehaviour

### LINKED ISSUE #3133: BadBehaviour.php has backtick in SQL, this isn't valid in PostgreSQL

tinoest:
Fixes #3133


emanuele45:
Thanks!

PR #3138, opened 2018-03-19
[WIP] Search refactor

emanuele45:
Do not merge, I just want to see the comparison between this and the previous of scrutinizer.


emanuele45:
Okay, this is not really stable, but at least the standard, fulltext and custom indexes seem to return something.
Now I would like to try to write some tests before giving the green light to review.

emanuele45:
@Spuds do you happen to have any idea what the f**k is complaining about now?
:cry: 

https://travis-ci.org/elkarte/Elkarte/jobs/403933470
```
Writing lock file
Generating autoload files
PHP Parse error:  syntax error, unexpected 'function' (T_FUNCTION), expecting identifier (T_STRING) or \\ (T_NS_SEPARATOR) in /var/www/vendor/myclabs/deep-copy/src/DeepCopy/deep_copy.php on line 5
```
it's something that tries to install via composer:
https://github.com/myclabs/DeepCopy/blob/1.x/src/DeepCopy/deep_copy.php
it doesn't like the:
```
use function function_exists;
```
that should be valid syntax as of php 5.6+
According to travis it should be using php 7.1.11 so in theory it should be working... :confused:

Spuds:
Seems this was introduced here: https://github.com/myclabs/DeepCopy/pull/120

We can try telling composer to load  the1.8.0 version instead.  In the composer.json file add
```"myclabs/deep-copy": "1.8.0"``` under the "require-dev" area and see it that fixes the error.  

Or possibly updating to the phpunit 6.5 branch ... right now we are on the 6.4.4.  I "locked: it there since that version and selenium were happy together so our webtests worked.  Not to say 6.5.x will not be fine as well, a quick test is all it takes to find out.

emanuele45:
Thanks @Spuds looks like it worked! :D

emanuele45:
YAY!
IT WORKS! :dancer: 

Well, at least the few tests I added, but that's already something.
The only problem still pending is that the test is run twice, so I have to add some hacks to the tests to avoid errors... :-\

Oh well, I think this could be considered done.
I have also done much more than intended in this PR, so I'm not sure if it is still relevant or not...
Have a look at it and let me know what you think.

The original idea was to make the API more "straightforward", going through a set of calls.
Then I added a bunch of stuff related to extracting the preparation of the messages for display (i.e. the callbacks), this part I'm not sure is really that relevant, but it should not harm. I'm not sure.
The search is still very complex, because there are a lot of parameters I'm not entirely sure how they interact and so I avoided removing them now. But, if you see anything that can be safely removed, feel free to point it out.

emanuele45:
Another note: this PR contains #3136 so the latter should be merged first (I think).

PR #3149, opened 2018-03-25
Fix/3141

### LINKED ISSUE #3141: Quick moderation "checkboxes" broken in development
When quick-moderation is enabled as "checkboxes", the following error stops js execution:
```
Uncaught TypeError: oButtonStrip.getElementById is not a function
    at InTopicModeration.handleClick (topic.js?R111:772)
    at InTopicModeration.init (topic.js?R111:694)
    at new InTopicModeration (topic.js?R111:679)
    at index.php?topic=547.0:1985
```

Additionally, the message ```Are you sure you want to do this?``` from the moderation actions is repeatedly shown while the page load.
It looks like 997f597f8be3916b3128a9d631aa5b21ee2b80b6 is the problem:
```
+				aEvents: [
+					['click', this.handleSubmit('remove').bind(this)
+				]
```
instead of attaching the event, ```handleSubmit``` is executed the moment the file is loaded.

@live627 any idea how to fix it?

live627:
Fix #3141 

> delegates

Yes, that's from C#.

### DIFF
--- a/themes/default/scripts/topic.js
+++ b/themes/default/scripts/topic.js
@@ -769,20 +769,20 @@ InTopicModeration.prototype.handleClick = function(oCheckbox)
 	// Show the number of messages selected in each of the buttons.
 	if (this.opt.bCanRemove && !this.opt.bUseImageButton)
 	{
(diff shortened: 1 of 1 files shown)


emanuele45:
Moar PHP less C#!! :P
Thanks!

PR #3154, opened 2018-03-27
Dealing with early errors

emanuele45:
There is still another potential issue here:
https://github.com/elkarte/Elkarte/blob/1c2044d6c6f58f6f5f9b713c9872c8a56bd0de4d/sources/Request.php#L425-L430

I don't think "http" in the query is **so** evil, provided the parser doesn't have other flaws.

Do you think it should be kept or it can be dropped?

### DIFF
--- a/sources/Load.php
+++ b/sources/Load.php
@@ -1830,6 +1830,11 @@ function loadEssentialThemeData()
 {
 	global $settings, $modSettings, $mbname, $context;
 
+	if (function_exists('database') === false)
+	{
+		throw new \Exception('');
+	}
+
 	$db = database();
 
 	// Get all the default theme variables.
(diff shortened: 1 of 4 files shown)


Spuds:
Would it make sense to split the string at that point and just keep the right or left side ?

emanuele45:
hmm... actually, I misunderstood the code... as usual. :laughing: 
This blocks only cases like:
```
https://www.elkarte.net/community/index.php?http://another_domain.tld/images/an_image.png
```
while I was thinking at things like:
```
https://www.elkarte.net/community/index.php?action=proxy;url=http://another_domain.tld/images/an_image.png
```

so I'd say it's not a big deal, it can stay.

PR #3157, opened 2018-04-04
And here it is: allow using the keys as substitution parameter for radio and select custom fields

emanuele45:
I put this together in a very not linear way, I started implementing it as key/val pairs stored in encoded json strings, then I moved to use the numeri key of the array.
I hope I cleaned up everything, but I'm not terribly sure of it, so anyone reviewing it: do it twice. :laughing: 

The upgrade doesn't fix the existing fields in order not to break anything that potentially already exist (i.e. if it is broken, it's easy to fix, if we broke it now it's worse).

### DIFF
--- a/install/install_1-1.php
+++ b/install/install_1-1.php
@@ -815,7 +815,7 @@ class InstallInstructions_install_1_1
 					'bbc' => 0,
 					'can_search' => 0,
 					'default_value' => 'undisclosed',
-					'enclose' => '<i class="icon i-{INPUT}" title="{INPUT}"><s>{INPUT}</s></i>',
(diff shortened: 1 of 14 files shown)


Vekseid:
I guess the only potential issue with this method is the difficulty of reordering.

But certainly fairly clever. : )

emanuele45:
Yep, the reorder can cause issues.
For 1.1 could be a stop-gap, and I think it's the less invasive one, for 2.0 we have time to either think a better solution or forget the issue. :laughing:

emanuele45:
Oh well, no comments it's good enough for me. :grin:

PR #3164, opened 2018-04-29
Fix Not Properly Merged

ahrasis:
Fix to change from DirectoryIterator to FilesystemIterator in Admin.subs.php was not properly merged and included in 1.1.3, both patch and install

This PR is to ensure that this should be included in 1.1.4 both patch and install.


ahrasis:
Please reject and delete this PR.

Issue #3120, opened 2018-03-05
Check all files against current versions error


ahrasis:
Opening an issue as discussed in the [forum](https://www.elkarte.net/community/index.php?topic=5091.0). A fix is also suggested in that discussion at [reply 6](https://www.elkarte.net/community/index.php?topic=5091.msg36138#msg36138).

emanuele45:
https://github.com/elkarte/Elkarte/pull/3121/commits/7be8ed961e1b6dd1d018b0fd1f8f5a3067f478c1

emanuele45:
#3166 that contains the above mentioned fix was sent to development instead of the 1.1 patch at the time, so the fix was missed in release.
This is still a valid bug for the 1.1 line.

PR #3165, opened 2018-04-29
Missing Fix For #3120

ahrasis:
Discussion and suggestion as in here: https://www.elkarte.net/community/index.php?topic=5091.0

But as noted, this was left out / missed in 1.1.3 both patch and install packages.

So, I am PRing this again.


ahrasis:
Please reject and delete this PR.

PR #3200, opened 2018-06-03
Last minute fixes

emanuele45:


Spuds:
Merged ... do we want to run 1.1.4 live for a few days as well ?

emanuele45:
I guess it would make sense...

PR #3229, opened 2018-07-28
Moar updates for 115

Spuds:
Its all about more, so here is some.


emanuele45:
Argh!
Forgot the header catchup in my branch... :sob: 
I wonder if it is possible to write a test ti check it. It should be doable I think :grimacing:

Issue #3225, opened 2018-07-13
Unexpected redirect to upgrade after install


emanuele45:
If the install directory is not deleted the just installed forum redirects to upgrade.
This is not technically wrong, because index.php finds the directory, and since there are correct data in Settings.php it *thinks* the install needs an upgrade.
Some more logic may be needed somewhere.

I would consider a "double redirect" to avoid too much logic in index.php:
1. in index.php have only a redirect to the install directory,
2. then move any more logic regarding what to do there (e.g. in an install/index.php)

Vekseid:
Suggestion

- Detect if upgrade file exists  
- If so, require version file in upgrade path.  
- If versions match, no redirect, warn admin.  
- If versions don't match, no redirect, link admin to upgrade.

PR #3242, opened 2018-09-16
Potential fix for #3225 checks to make sure either install.php or upgrade.php exists before redirecting

tinoest:
Added a additional check to ensure that install.php or upgrade.php exist in the install directory before redirecting the user to there.


tinoest:
@emanuele45 I don't disagree, this only meets the first part of that set of checks. i.e. it checks to see if the install directory has a upgrade.php inside it. 

I'm not sure what is bad about leaving the install directory lying around and being empty, but I can add a warning when your on the board index to state it still exists and should be removed?

PR #3256, opened 2018-11-09
Rework of namespaces, directory structure and class loader

### LINKED ISSUE #3253: Naming conventions (again)
I was thinking what would you @Spuds think if for 2.0 we make all the naming of classes and files a bit more consistent?
It will likely break quite a bit of stuff, but at least it will make (for example) the autoloader much simpler.

With more consistent I mean:
1) drop the ".class", ".interface", ".abstract" (if that one was ever used), etc. bits from the file name,
2) make the namespaces consistent (I'm not yet sure how that could be, but I think something like in "sources" all "ElkArte", in "subs"... I have no idea, etc. :laughing: )
3) drop the "_controller" from class names (and use a namespace instead)
4) have abstract classes consistently having "Abstract" at the beginning of the name,
5) have the interfaces consistently having "Interface" at the end of the name (just because all of them right now have it that way lol)
etc.

emanuele45:
I used the composer classloader (PSR4), on top of that moved all the classes to an ElkArte directory (to which an ElkArte namespace corresponds) inside sources. From there reworked all the namespaces and class names and file names in order to match the convention.

Fixes #3253


emanuele45:
Codacity instead is always complaining! :rofl:

emanuele45:
BTW, I forgot I wanted to tackle also the database classes!! :man_facepalming:

Spuds:
I'll wait to merge it then ... so far it really looks awesome

emanuele45:
Not yet working, but just to give any idea of what I had in mind (related also to #3257 (not yet) and #3258).

Spuds:
Looking good !   Would the plan be to update our db calls so we can in the core not use any of the temporary migration functions and just leave those for addons?

emanuele45:
Well... I doubt I can update all the calls in this very PR, though I added a backward compatibility "layer" with tracked errors, so that old-style calls can be identified.

Spuds:
Cool .. yes I meant change them as we go along, not all in this PR, that would be crazy man !

emanuele45:
I'm almost lost... can't understand what the heck it fails on installing the tests with an
> Fatal error: Allowed memory size of 1073741824 bytes exhausted (tried to allocate 36864 bytes) in /var/www/sources/ElkArte/Database/Mysqli/Connection.php on line 43

:confounded:

emanuele45:
YAY!
Finally tests are green again!

One thing I'm not sure about is the upgrade, but to be honest, at the moment I don't want to know. :laughing:

Spuds:
MERGE !!!!!!

emanuele45:
:dancing_women: :dancing_men: :dancer: :dancing_men: :dancing_women:

PR #3261, opened 2018-11-24
Dropped "db_" prefix to many database-related methods and some more

emanuele45:
* added a fetch_all
* removed the "db_" thingy from method names
* removed some static variables
* simplified the "query" code moving out common stuff between mysql and postgre
* removed a begin/end transaction happening in query based on some kind of random guesses based on the assumption that if transaction is necessary it's up to the code writer to initiate and terminate it.


Spuds:
Be nice to get rid of all those db_ 👍

emanuele45:
So, after all this crap, I have no idea why it doesn't work. :joy: 
Apparently, for some inexplicable reason, the insert of the settings table during install for postrgesql doesn't insert anything. And from there everything else fails.
I have to dig more into it... :-\

emanuele45:
And finally, after so many tries I even lost count, the tests are finally back working!!
 :dancing_men: :dancing_women: :dancer: :dancing_women: :dancing_men:

emanuele45:
I think i don't want to touch this any further... and I hope I didn't left around any debug code... :sob:

Spuds:
I'm going to merge it in so you avoid the temptation of adding to it !!! ... Really awesome work here 👍

PR #3271, opened 2018-12-25
A few misc items fixed

Spuds:


sources/ElkArte/Controller/Register.php
```diff
@@ -237,8 +237,8 @@ public function action_register()
 			{
 				// No file found or a blank file, log the error so the admin knows there is a problem!
 				loadLanguage('Errors');
-				Errors::instance()->log_error($txt['registration_privacy_policy_missing'], 'critical');
-				throw new Elk_Exception('registration_disabled', false);
+				\Errors::instance()->log_error($txt['registration_privacy_policy_missing'], 'critical');
```
emanuele45: I'm frequently confused by namespaces... (that's why I tried to always use the fully qualified notation :laughing: ).
So even if Errors is in the ElkArte namespace here it needs only the "root" \?
Good to know, I hope to remember.

emanuele45:
ohhh come on! There is nothing like "too many!! :laughing: 

Great cleanup, thanks!!

sources/ElkArte/Controller/Register.php
Spuds: Me too, so I hope I got most of them right.

PR #3272, opened 2018-12-27
Reduce globals 2

emanuele45:
Another attempt at removing some globals.
I started somehow easy with $boardList, $boards, $cat_tree, $fn_count and $fn_num, then I got carried away by $gd2 and moved quite a big of stuff around...
@Spuds what do you think? The images code is your field, I never touched it much... :fearful:


Spuds:
LOL ... I just started doing something along the same lines ... 
Creating an abstract image class and then separate GD and Imagick class to extend that ... basically separate the spaghetti code of GD vs Imagick.  

Was also going to break up some of the functions since they call one another from inside the functions vs proper chaining in the calling functions.

So I love the idea, and you are farther along then I at this point, go man go !

emanuele45:
NOOOO!!!! :laughing: 

Then pick pick whatever you want from this and I continue with something else! :D

Spuds:
OK I'll continue onward and pick any goodies that I see.

emanuele45:
Oook.
I'll close that one, leave the branch there for reference and create a new one without this last commit?

Spuds:
That will work, I actually pulled it into a temp branch on my local so I have a copy.  ... On my current test forum for some reason the attachments area is not showing up under the posting area?  Any ideas?  My current test forum is not a clean install but a upgraded 1.0 so don't know if its missing something in the db.

emanuele45:
From this branch? Or in general with development? (Sorry, I didn't test it that much recently and anyway I have a messy install.)

Spuds:
Latest dev branch ... when I go to post I don't see the attachment area, (there is a poll undefined index in there as well)  Glad you also have a messy install, helps to find bugs !

Issue #3193, opened 2018-05-28
Rework attachments drectory/ies settings


emanuele45:
The first and main proposal is to drop the support for the single directory option and have a default multi one.
I would go with the year/month pair, that should be enough in the vast majority of the cases.

The second is drop the serialized structure in favor of the simpler json.

Also the code requires some reorganization, but I guess this will come while working.

Spuds:
😸

Vekseid:
My suggestion would be to have hexadecimal hash subdirectories, rather than by date. 

json rules phpserialize drools.

PR #3275, opened 2018-12-29
Removal of another few globals and cleanup

emanuele45:


Spuds:
498 files changed 🥇 Awesome cleanup.

emanuele45:
Well... mostly was the changes in the headers! :laughing: 
Okay, now I think I'll have another look at the attachments code, I realized there is something that smells there... :-\

Spuds:
I think a lot of whats going on is the new name spacing ... the ::hooks functions in the various modules need to be fully qualified.  Also at least for attachments it calls off Attachments_Post_Module which is now just Post.   At least I think that is the base of the issue of why it (and drafts etc) are not loading.

emanuele45:
hmm... they should be already fully qualified, maybe I changed the names and didn't update them in upgrade or install... that could very well be the reason.
Up to 1.1 the name of the class was something like Attachments_Post_Module, while in dev is Post.

Spuds:
That takes care of it... FYI I had changed the code in the getModuleClass to be
```
		$module_class = explode('\\', trim(get_class($this), '\\'));
		$module_class= end($module_class);
```
I thought that was more generic, but the str_replace is fine as well.  You may need to rebase to avoid the conflict.

emanuele45:
It does make sense. Since it's looking for the controller name it should work much better! :+1:

emanuele45:
BTW I didn't apply the patch, just generated it, so feel free to apply it, I'll take it in when is merged. ;D

Spuds:
Will do .. I have it as part of my graphic.subs change over to a class branch.  I'm getting there, but  it will need more work as unraveling the madness was a bit messy 💃

emanuele45:
LOL
I've worked a bit on the attachments directories now to streamline madness there.
And I'll start checking also attachments themselves to see if I can get something more... manageable.hash('sha1', hash('md5', $filename . time()) . mt_rand())

Spuds:
We are only going with auto years on the automanage directory thing correct ??  At least I recall a discussion along those lines.  Removing all those options should be a help.

Maybe a good use for the TokenHash class that we have, it should be able to generate any hash length that we need.

emanuele45:
There is #3193 about it, yes.
For the moment, though, I'm trying to actually find the logic behind all the juggling that an attachment does. :rofl:
This is where I am right now:
https://github.com/elkarte/Elkarte/compare/development...emanuele45:attachments_rework?expand=1

PR #3279, opened 2019-01-05
Plus115

Spuds:
As near as I can tell these commits from 1.1.5 were not correctly merged into development, so here there are.


emanuele45:
/me is mostly innocent... I swear! :joy_cat:

PR #3280, opened 2019-01-10
Beginning of Graphics restructure

Spuds:
Still needs work but its a start !


sources/ElkArte/Graphics/Gd2.php
```diff
@@ -0,0 +1,494 @@
...
+	/**
+	 * Output the image resource to a file in a chosen format
+	 *
+	 * @param string $file_name where to save the image, if null output to screen
+	 * @param int $preferred_format jpg,png,gif, etc
+	 * @param int $quality the jpg image quality
+	 *
+	 * @return bool
+	 */
+	public function output($file_name = null, $preferred_format = IMAGETYPE_JPEG, $quality = 85)
```
emanuele45: Scrutinizer says it should be string|null, though maybe the default should be an empty string because of the strict comparison afterwards?

sources/ElkArte/Graphics/AbstractManipulator.php
```diff
@@ -0,0 +1,178 @@
...
+	 * @return mixed
+	 */
+	abstract public function generateTextImage($text, $width = 100, $height = 75, $format = 'png');
+
+	/**
+	 * Loads a image file into the image engine for processing
+	 *
+	 * @return bool|mixed
+	 */
+	abstract function createImageFromFile();
```
emanuele45: scrutinizer says the scope is missing. :angel: 
Same for other abstract in this class.

sources/ElkArte/Graphics/Gd2.php
```diff
@@ -0,0 +1,494 @@
...
+
+/**
+ * Class Gd2
+ *
+ * @package ElkArte\Graphics
+ */
+class Gd2 extends AbstractManipulator
+{
+	/** @var resource */
+	protected $_image;
```
emanuele45: This should be declared in the AbstractManipulator class, otherwise scrutinizer complains that here https://github.com/elkarte/Elkarte/pull/3280/files#diff-d62ebe3b9dd825fc67c0d66effaeba72R119 the property ->_image is not defined. :D

sources/ElkArte/Graphics/Gd2.php
```diff
@@ -0,0 +1,494 @@
...
+
+	/**
+	 * Loads an image from a web address into the image engine for processing
+	 *
+	 * @return bool|mixed
+	 */
+	public function createImageFromWeb()
+	{
+		require_once(SUBSDIR . '/Package.subs.php');
+		$this->_image = fetch_web_data($this->_fileName);
```
emanuele45: $this->_image should be $image I think

sources/ElkArte/Graphics/Imagick.php
```diff
@@ -0,0 +1,398 @@
...
+
+	/**
+	 * Loads an image from a web address into the image engine for processing
+	 *
+	 * @return bool|mixed
+	 */
+	public function createImageFromWeb()
+	{
+		require_once(SUBSDIR . '/Package.subs.php');
+		$this->_image = fetch_web_data($this->_fileName);
```
emanuele45: Shouldn't be $blob = fetch_?

sources/ElkArte/Graphics/Gd2.php
Spuds: null outputs to the stream to the buffer, I did that vs '', my intention was for an implicit request vs just a mistake,  

I'll add the string|null BUT I bet Scrutinizer will complain about the other called methods in output that use file_name but require a string .... So I guess to clear that "error"  I'll have to drop the null and just use a blank string file_name to echo the blob.

sources/ElkArte/Graphics/AbstractManipulator.php
Spuds: Yup ... Guilty

sources/ElkArte/Graphics/Gd2.php
Spuds: I'll add it there and see how it goes ... I moved it to the specific manipulators since $_image can be either a gd resource or Imagick one.   If you only declare it in the abstract as both (\Imagick|resource) then Storm lights up with unknown method errors, so I'm going to leave them in the manipulators as well.

sources/ElkArte/Graphics/Gd2.php
Spuds: Its a cheat  .... $this->getSize expects its data to be available in $this->_image, so although at that point its not an image but a string,  I did not want to pass the value.

sources/ElkArte/Graphics/Imagick.php
Spuds: As above ... its a result of $this->getSize('string'); wanting $this->_image to be populated.  And as above to fix would need to pass the value instead.

Spuds:
Lets see what Scrutinizer has to say about that.  I'm actually not sure if  GD's imageXXX functions will echo the output when the filename is '', the docs seem to imply that it must be null ... If thats the case some '' => null conversion will need to be done in the GD manipulator.  Imagick works differently so should be fine.

Spuds:
Why the conflict now ??? oh come on, I'm sick 😷 and don't want to play around with git right now.

PR #3283, opened 2019-01-16
Update several EXT libraries

Spuds:
Lots-o-updates to the libraries ... should make things more 7.x compliant


emanuele45:
Bring it in! Bring it in! :dancer:

PR #3289, opened 2019-03-13
Some necessary fixes for code I broke "recently"

emanuele45:


emanuele45:
@Spuds I guess we can apply the "merge early and deal with broken stuff" here too, right? :rofl:

Spuds:
Yup ... I was doing the same thing as you this PM but you beat me to it :D ... the other error in fetch web stream is a legit bug that I have fixed in my repo .... PR soon 😇

PR #3303, opened 2019-04-11
Update Session.php to secure PHPSESSIONID

ahrasis:
This is a suggested update to secure PHPSESSIONID as discussed in https://www.elkarte.net/community/index.php?topic=5596.0


Spuds:
@emanuele45 what do you think ?  Request was [here](https://www.elkarte.net/community/index.php?topic=5596)

emanuele45:
I'm not really a session expert... :-\
@ahrasis did you test the registration works? This is usually the most critical part. If this has no problems I think it should be fine.

ahrasis:
I tested this on a freshly installed EA forum and registration works without any problem.

PR #3304, opened 2019-04-12
The two "security" issues open

emanuele45:
@Spuds please have a look at the search part, I *think* it should work, but with search you never know. :laughing:


Spuds:
That looks good to me, and from a couple of test cases is doing what it should.

Since you are changing the loadMessagesRequest function you could also inner joint => join there like you did elsewhere in this PR.

PR #3318, opened 2019-06-05
I don't think there files actually changed ?

Spuds:
I don't see any changes to these files since 1.1.5, so this reverts the headers back.  Am I missing a change somewhere?


emanuele45:
Darn!
Master was at 1.1.4, I updated locally the repo I use for texting, but not the one I use for releases... :man_facepalming:

PR #3319, opened 2019-06-07
Updates to permissions class

Spuds:
Sorry for the last second bonus fixes, but there were some soft errors to clean up.


emanuele45:
I'd say there is nothing left but prepare the packages!

PR #3327, opened 2019-07-24
Update to 1 1 6

emanuele45:


emanuele45:
As usual, @Spuds give a quick review, this kind of merges are always rather tricky...

Spuds:
"tricky" is the nicest description I can think of ... most of the Travis errors came from curl fetch web so those are somewhat bogus .... There was one other that I have not looked at.

emanuele45:
Well, if it is just the post I'd be happy. :joy:
I meant more in general due to the number of conflicts that are easy to overlook... :-\

Spuds:
I'm going to pull this in, we can deal with the aftermath then 😇

PR #3331, opened 2019-08-12
[WIP] Moving $user_info and $user_settings out of global (static class for now)

emanuele45:
In the last couple of days I got bored and started playing with this.
Basically for now is just replacing two globals with a a pair of static variables in a class (made of two objects ValuesContainer).
$user_settings is "read-only" because I really never understood it's meaning and to me it looks like the only reason for it to exist is to be a read-only copy of the data pulled from the db.
Doing that I noticed some few exceptions, and I still have to deal with them somehow (that's why this is still a WIP), TBH I'm not really sure how exactly, so ideas are welcome! :laughing: 
$user_info for the time being is a mirror of ElkArte\User::$info so it still work (replacing all of them requires a bit of attention).
The forum loads bit I'm pretty sure there are tons of errors.

TBH I did this just because I felt, no real reason. So if you see too many troubles in this direction let me know and we can scrap it. :smile:


Spuds:
I kind of like this direction ... I'd say go for it

Spuds:
I must say this may be a record LOL  ```Tests: 202, Assertions: 36, Errors: 187, Failures: 3, Warnings: 1, Skipped: 2.``` ... something is making phpunit some crazy with the errors.

emanuele45:
```
Exception: Serialization of 'class@anonymous' is not allowed
```
:roll_eyes:

Spuds:
I don't know if maybe updating phpunit from v6 to v7  would help ... even v8 if we  only test php 7.2+

emanuele45:
Oh well, now the tests are fixed! :D 
@Spuds can you have a look at 491f4d51a0c8132704d595d2d6ba8189ce105c8f I tossed it in because I needed to fix some other issue I was having, though I'm not sure what I'm doing there... :blush:

emanuele45:
I was planning to cleanup all the $user_info, but I'll do in another PR! :D

Spuds:
I figured since the tests were at least passing it was time to do the first pull in :D

PR #3332, opened 2019-08-14
Travis testbed updates and other fixes

Spuds:
First this moves from trusty to xenial for the tests, this mainly due to trusty's issues with getting mysql 5.7 installed and since thats the default for xenial ....

Fixed a few legit errors in the code that the tests were failing on

Made some changes so jshint will give us a clean pass, I *did* supress 2 errors with todo's in the code.  There were 2 functions inside of loops that I just did not want to deal with right now.

So this passes on my local, lets see what it does here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 sudo: required
 language: php
-dist: trusty
+dist: xenial
 
 matrix:
   include:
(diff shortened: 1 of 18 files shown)


Spuds:
Ohhh a pretty green checkmark 🎉

PR #3334, opened 2019-08-16
Fix a few annoyances in the package manager

Spuds:
Couple of minor bugs, couple of minor tweaks.


themes/default/Packages.template.php
```diff
@@ -500,11 +499,11 @@ function template_examine()
...
 	echo '
 	<div id="admincenter">
-		<h2 class="category_header">', $txt['package_examine_file'], '</h2>
-		<h2 class="category_header">', $txt['package_file_contents'], ' ', $context['filename'], ':</h2>
-		<div class="content">
-			<pre class="file_content">', $context['filedata'], '</pre>
-			<a href="', $scripturl, '?action=admin;area=packages;sa=list;package=', $context['package'], '">[ ', $txt['list_files'], ' ]</a>
+		<h2 class="category_header">', $txt['package_examine_file'], ' : ' , $context['package'], '</h2>
+		<h3 class="category_header">', $txt['package_file_contents'], ' ', $context['filename'], ':</h3>
+		<div class="content" style="font-size: 1.1em">
```
emanuele45: Is that inline intended?

themes/default/Packages.template.php
```diff
@@ -746,211 +745,19 @@ function template_file_permissions()
...
-			0: "read",
-			1: "writable",
-			2: "execute",
-			3: "custom",
-			4: "no_change"
-		}
+		var package_file_perms_writable = "', $txt['package_file_perms_writable'], '",
+			package_file_perms_chmod = "',  $txt['package_file_perms_chmod'], '",
+	    	package_file_perms_not_writable = "', $txt['package_file_perms_not_writable'], '",
+	    	package_file_perms_more_files = "', $txt['package_file_perms_more_files'], '";
```
emanuele45: JavaScriptEscape?

themes/default/Packages.template.php
Spuds: Of course ... I mean you don't think I would be lazy and just add stuff like that do you ??? 😄

themes/default/Packages.template.php
Spuds: Indeed

themes/default/Packages.template.php
emanuele45: Absolutely not. :joy_cat:

PR #3336, opened 2019-08-21
Ci tweaks

Spuds:
Some cleanup and updates, 
 - newer version of selenium, 
 - turned off opcache for the tests
 - move webtests to be standalone suite
 - use phpunit --prepend to improve coverage test speed
 - add 7,2 to the mix,
 - move db and bootstrap tests as testsuites to try to catch all test coverage
 - fix postgre issues with number of values in the settings table (well fix is a bit strong here) also updated install code to more closely match the mysql version


Spuds:
Code coverage % has gone down, although covered lines has gone up.  The total number of lines is now more honest  (still needs some tweaks but ....) so the % is more accurate.  i.e. the truth hurts.

emanuele45:
hmm... how the heck do they count covered lines?
I mean, if you remove lines from the total the % shouldn't go up?

Spuds:
With this change it "merges" the code coverage from 3 tests, mysql, postgre and webtest.   When I did this I added the files under theme/default and that added to the total lines in the project.  That increase was far greater than the coverage we have on them ;(

Previously it was 10,515 Covered Line out of  82,050  Total Lines
Now its 11,480 Covered Lines but 126,469 Total Lines

I'll play around with processUncoveredFilesFromWhitelist and addUncoveredFilesFromWhitelist just for the webtests and see if that helps the number, even if only artificially.

Spuds:
Ha ... that little tweak changed it to 
11,480 Covered Lines, 82,850 Total Lines, 13.9 % Coverage

So for the postgre and webtests it only adds to the total lines for files that are being tested.  The mysql test adds "all" the files regardless.  Really this just prevents adding in "all" the template files and instead only adds in the ones currently under test.

emanuele45:
So this is can be merged?

Spuds:
Bwahahahahh yes

PR #3338, opened 2019-08-25
Removed $user_info from all the sources/ElkArte path

emanuele45:
Except for few that are still necessary to keep backward compatibility.
I also removed the 1.1.x patches left around by mistake during my last update.

@spuds please double check 165563dba094279cbff6cc455709fa9f4d3f358f I *think* it's okay, but I'm not entirely sure. It was a mistaken removal during one of the merges from master... :-\


Spuds:
That looks good to me

emanuele45:
Okay, that one is done as well.

PR #3347, opened 2019-09-21
Healing Patch update and code trolling

Spuds:
A couple of minor things needed tweaks for the user_info updates

While I'm at it, I've begun a code trolling :trollface: exercise :D ... removing unnecessary if/else statements or logic, move some ancient for () style constructs to foreeach, some more == to === and stuff like that.

OK since this is still open, lets get a bit crazy ... remove phpass (PasswordHash.php) since native php functions in 7.0+ (maybe 5.5+) should do exactly the same thing, meaning removing the lib and using the built in should be "transparent" and just work tm see https://github.com/elkarte/Elkarte/pull/3347/commits/9b6118ba9bdecab0c3ac59b8f7ada5195f11597f

While I'm at it, I moved most createlist callable array syntax to anonymous functions calls, so ```array($this, 'function_name')``` goes to ```function ($start, $items_per_page, $sort) {return $this->function_name($start, $items_per_page, $sort);}```


sources/ElkArte/AdminController/Maintenance.php
```diff
@@ -1051,7 +1051,7 @@ public function action_backup_display()
 		checkSession('post');
 
 		// Validate access
-		if (defined('I_KNOW_IT_MAY_BE_UNSAFE') === false && $this->_validate_access() === false)
+		if (!defined('I_KNOW_IT_MAY_BE_UNSAFE') && $this->_validate_access() === false)
```
emanuele45: Just for consistency's sake, _recently_ I've been using ```=== false``` in this type of comparisons mostly because from time to time the ```!``` seems to me less easy to catch.
I don't really mind one way or the other, so should we always use the ```!```?

Spuds: I don't think it matters that much, but since when we are already dealing with a boolean value why do another logical compare to get a boolean value again?  I think using the ! with a well named variable makes for cleaner code.

sources/ElkArte/AdminController/ManageThemes.php
```diff
@@ -910,10 +909,12 @@ public function action_remove_api()
 
 		// It is a theme we know about?
 		$known = explode(',', $modSettings['knownThemes']);
-		for ($i = 0, $n = count($known); $i < $n; $i++)
+		foreach ($known as $i => $known)
```
interlab: typo?

Spuds: Good catch, thanks !

interlab:
https://github.com/elkarte/Elkarte/blob/master/CONTRIBUTING.md

https://gist.github.com/3342247

404 Page not found

Spuds:
I *think* that is now https://gist.github.com/piscisaureus/3342247 ... have to update the MD

Spuds:
@emanuele45 since this is a bit of a cleanup PR (now at 200+ files), do you want me to do the dreaded {} formatting update we said we would follow?  I'd just batch it off so its easy and finally over with.  If not I can do it in another PR or we can just attempt to let it happen organically as we continue with updates?

emanuele45:
If you want go for it!
/me hides

Spuds:
@emanuele45 could you take a look at https://github.com/elkarte/Elkarte/pull/3347/commits/f097f0e5c30931e4311b382995a143d419a9a7c9 When I was in the ACP and went to modify boards it showed only the category names with no boards under them.  This change fixes that but I'm not sure if there are other issues or if this was the best fix.  Thanks!

emanuele45:
hmm... the original was:
```php
/**
 * Recursively get a list of boards.
 *
 * - Used by getBoardTree
 *
 * @package Boards
 * @param int[] $_boardList The board list
 * @param array $_tree the board tree
 */
function recursiveBoards(&$_boardList, &$_tree)
{
	if (empty($_tree['children']))
		return;

	foreach ($_tree['children'] as $id => $node)
	{
		$_boardList[] = $id;
		recursiveBoards($_boardList, $node);
	}
}
```
so I guess it makes sense indeed.

Spuds:
> If you want go for it!
> /me hides

You asked for it .... but I get scruit now thinks I created all the issues!

emanuele45:
Spuds wrote on 29/09/2019 16:42:
>
>     If you want go for it!
>     /me hides
>
> You asked for it .... but I get scruit now thinks I created all the 
> issues!
>

You broke everything!!!11!1!

emanuele45:
/me looks at the name of the branch, then at the number of rows changes, then back at the name of the branch... :see_no_evil:

Spuds:
> /me looks at the name of the branch, then at the number of rows changes, then back at the name of the branch... 🙈
> 
> @Spuds when this is ready I guess you can just merge it. 😹

It really did start out as minor changes LOL ... but I kept finding things and then wanted to use some 7.x goodness, and the formatting had to be done.

Really it was not my fault, an old friend came in from out of town, my car broke down, I had no money for a taxi, there was an earthquake, a terrible fire, locusts! .... 

hey scruit says ```694 issues were fixed``` and that is not even the real bugs that got stomped!

I'm going to look over the templates directory this PM, and then merge this 🐷 in so branch work can be done again.

Spuds:
And there it is .... plop.

PR #3352, opened 2019-10-10
$this->_req tweaks

Spuds:
This is mostly some tweaks to the http req class to add a compare function (post and get) as well as a request style function which looks at get then post.

Implemented and cleaned up some existing uses, mostly in the admin directory, of the above

Made some changes to the action controller so it should be even more robust, or I broke it, its one of the two.

Found a few bugs along the way and fixed those as well as some spelling and misc formatting just cuz

### DIFF
--- a/sources/ElkArte/Action.php
+++ b/sources/ElkArte/Action.php
@@ -61,42 +61,42 @@ class Action
 	}
 
 	/**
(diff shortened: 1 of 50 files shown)


emanuele45:
I missed that one... well, let's not waste more time! :D

PR #3353, opened 2019-10-13
Just a test of testing

Spuds:
Just playing with trying to get controller testing done since there is no browser, login, session, etc.  Want to see what this does to the coverage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,9 @@ dist: xenial
 
 matrix:
   include:
-    - php: 7.1
+    - php: 7.3
       env:
-        - DB=mysql-5.7
+        - DB=postgres-9.6
         - WEBTESTS=false
         - COVERAGE=true
     - php: 7.3
@@ -46,7 +46,6 @@ matrix:
 
 services:
   - memcached
-  - mysql
 
 # Prepare the server (PHP, Apache, PHPUnit, APCu etc)
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,39 +9,6 @@ matrix:
         - DB=mysql-5.7
         - WEBTESTS=false
         - COVERAGE=true
-    - php: 7.3
-      env:
-        - DB=mysql-5.7
-        - WEBTESTS=true
-        - COVERAGE=true
-    - php: 7.3
-      env:
-        - DB=postgres-9.6
-        - WEBTESTS=false
-        - COVERAGE=true
-      addons:
-        postgresql: "9.6"
-    - php: 7.2
-      env:
-        - DB=mariadb-10.3
-        - WEBTESTS=false
-        - COVERAGE=false
-      addons:
-        mariadb: '10.3'
-    - language: node_js
-      node_js:
-        - "node"
-      php:
-      # none
-      install:
-        - npm install --save-dev jshint stylelint stylelint-order stylelint-config-recommended
-      before_script:
-      # nothing
-      script:
-        - node_modules/jshint/bin/jshint .
-        - stylelint "**/*.css"
-      after_script:
-      # nothing
   fast_finish: true
 
 services:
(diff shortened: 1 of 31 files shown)


Spuds:
```The last posting from your IP was less than 0 seconds ago. Please try again later.``` WTF .7.3/postgre !
I tried both random IP's and zero waitime with the same fail.   @emanuele45 I assume this is a db abstraction error where its returning a value other than ```1``` for ```affected rows (security 1531)``` but I don't have any postgre setups to test on to fix this issue so could use some help.  

On another note, this simple test adds .8% to the coverage even though we test create post and likes already, but just the subs, not the controllers.

**Comment by emanuele45** [comment]
I'm looking into it.
Note that there is a bug in bootstrap.php:
```
if (defined('IGNORE_INSTALL_DIR'))
```
should be:
```
if (!defined('IGNORE_INSTALL_DIR'))
```

**Comment by emanuele45** [comment]
Anyway @Spuds I'm currently stuck with not being able to connect to the psql server. :-\
I'll keep trying...

**Comment by Spuds** [comment]
I did some abuse of Travis to help debug a bit.  What I *think* is occurring is that ```$this->_db_replace_result``` is being set in the PG insert function (part of the pgsql replace compatibility area) but after this section, right before the Result object is set
```
		if (!empty($data))
		{
			$last_inserted_id = $this->insert_id($table);
		}
```
the value of ```$this->_db_replace_result``` is reset to null, and that is what is returned in the Result object and then returned in ```affected_rows``` so the spam protection triggers at that point since it does not get a ```1``` back.

**Comment by Spuds** [comment]
Further looking has this section in the query function in Query.php (postgre)
```
		// One more query....
		$this->_query_count++;
		$this->_db_replace_result = null;
```
so this is where the results are nullified.  As a test I changed it to ```$this->_db_replace_result = $this->_db_replace_result ?? null;``` and the tests passes.  Not sure if that is the right thing, or changing the logic around in the insert function.

**Comment by emanuele45** [comment]
@Spuds I think the issue is with inser_id that does a query that should not be "counted" towards.
I think the easiest way should be to change:
```
$this->_db_replace_result = $db_replace_result + $inserted_results;
```
to:
```
$db_replace_result = $db_replace_result + $inserted_results;
```

and then:
```
'replaceResults' => $this->_db_replace_result,
```
to:
```
'replaceResults' => $db_replace_result ?? 0,
```

Here it works and **should** avoid side-effects. Maybe not all of them, though. :laughing:

Spuds:
Is anything then setting ``` $this->_db_replace_result``` or would that always be null ?

emanuele45:
@Spuds I guess it would not be set ever. Fair enough. :laughing: 
hmm... this is a pain in the arse... :-\
And actually, this is the only instance I could find that requires this hack in order to give "correct" results.
Considering how seldom (hopefully) this code would run, would it make sense to just have one more query and not rely on an insert replace?
Unless this could have implications also for example for SP.

Spuds:
Doesn't postgre have a ```ON CONFLICT``` statement you can use on insert?  Could we use that or is that different altogether.  I'd be fine with another query, I'm not sure of the implications on SP, its been a while since I looked at the code 😇

emanuele45:
Looking at the old SMF 2.1 code I have handy, this variable was never set to begin with... :thinking:

emanuele45:
@Spuds what is the status of this one? Should be merged?

Spuds:
Let it sit for a bit ... I don't have your suggested postgre db fix in this and have another test to push.  Got sidetracked on some 1.1.6 issues

Spuds:
I'm done playing with this for now, going to merge it in as the updates will allow us to do some controller testing w/o having selenium running.  Not perfect nor exhaustive but a start.

emanuele45:
:dancing_men: :dancing_women: :dancer: :man_dancing: :dancing_women: :dancing_men:

PR #3354, opened 2019-10-13
Consolidate attachments drectories handling

emanuele45:
I gave it a go at re working the attachments. I started with directories trying to move everything into a single class. Well I moved as much as possible, except several settings. But at least it gives an idea.
What do you think @Spuds ?


sources/ElkArte/AttachmentsDirectory.php
```diff
@@ -0,0 +1,876 @@
...
+	public function remainingFiles($current_files)
+	{
+		if ($this->hasNumFilesLimit())
+		{
+			return max($this->numFilesLimit - $current_files, 0);
+		}
+		else
+		{
+			return false;
+		}
```
Spuds: No need for the else, simply return false 😇

sources/ElkArte/AttachmentsDirectory.php
```diff
@@ -0,0 +1,876 @@
...
+	public function remainingSpace($current_dir_size)
+	{
+		if ($this->hasSizeLimit())
+		{
+			return max($this->sizeLimit - $current_dir_size, 0);
+		}
+		else
+		{
+			return false;
+		}
```
Spuds: Ditto, I've been on a remove unnecessary else statements bender

sources/ElkArte/AttachmentsDirectory.php
```diff
@@ -0,0 +1,876 @@
...
+	{
+		if (is_integer($id))
+		{
+			return array_key_exists($id, $this->attachmentUploadDir);
+		}
+		else
+		{
+			return in_array($id, $this->attachmentUploadDir) || in_array(BOARDDIR . DIRECTORY_SEPARATOR . $path, $this->attachmentUploadDir);
+		}
+	}
```
Spuds: Else's, we don't need no stinking else's!, just return the stuff 💃

sources/ElkArte/AttachmentsDirectory.php
```diff
@@ -0,0 +1,876 @@
...
+	public function isCurrentBaseDir($id)
+	{
+		if (is_integer($id))
+		{
+			return $this->basedirectory_for_attachments === $this->attachmentUploadDir[$id];
+		}
+		else
+		{
+			return $this->basedirectory_for_attachments === $id;
+		}
```
Spuds: What are you paid by the number of else's :trollface:

sources/ElkArte/AttachmentsDirectory.php
```diff
@@ -0,0 +1,876 @@
...
+	{
+		if (isset($this->attachmentUploadDir[$id]))
+		{
+			return $this->attachmentUploadDir[$id];
+		}
+		else
+		{
+			throw new Exception('dir_does_not_exists');
+		}
+	}
```
Spuds: same

sources/ElkArte/AttachmentsDirectory.php
```diff
@@ -0,0 +1,876 @@
...
+	{
+		if ($minLevel === null)
+		{
+			return !empty($this->automanage_attachments);
+		}
+		else
+		{
+			return $this->automanage_attachments > $minLevel;
+		}
+	}
```
Spuds: Else always return, so return

sources/ElkArte/AttachmentsDirectory.php
```diff
@@ -0,0 +1,876 @@
...
+		}
+		elseif (in_array($updir, $modSettings['attachmentUploadDir']))
+		{
+			$outputCreation = true;
+		}
+		else
+		{
+			$outputCreation = false;
+		}
+
```
Spuds: Since I'm being a total PITA, just set outputCreation to false before the first if and drop the last else, that else is just there to ensure var initialization, so initialize it to begin with.

Spuds:
Looks awesome! So good to get that code pulled out to its own area.  I made some persnickety comments that you can ignore, since I've been lacking any code content lately I've been going for code style points 💃

emanuele45:
:sunglasses: 
On the "else" I usually tend to prefer to be explicit to avoid saving 4 letter and risk later not to know why that return was put there. But I'm not that strict on it. I'll keep them at least until the code is fine (and at the moment I'm pretty sure it is not :laughing: ) so that in case of moving around stuff the logic is not lost.

emanuele45:
Ops... I guess we had a huge misunderstanding (my fault, I didn't put [WIP] in the title of the PR), this was just a "what do you think"-kind of PR.... :sob: 
Oh well, it just means I have to finish it up for real this time. :laughing:

Spuds:
Since you would not remove your else's I had to force your hand to finish the rest 👿 LOL

PR #3363, opened 2019-10-31
My117

Spuds:
Start to get some 1.1.7 items addressed

### DIFF
--- a/sources/controllers/Notify.controller.php
+++ b/sources/controllers/Notify.controller.php
@@ -29,6 +29,7 @@ class Notify_Controller extends Action_Controller
 	{
 		// Our topic functions are here
 		require_once(SUBSDIR . '/Topic.subs.php');
+		require_once(SUBSDIR . '/Boards.subs.php');
 	}
 
 	/**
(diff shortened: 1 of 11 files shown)


sources/controllers/Notify.controller.php
```diff
@@ -29,6 +29,7 @@ public function pre_dispatch()
 	{
 		// Our topic functions are here
 		require_once(SUBSDIR . '/Topic.subs.php');
+		require_once(SUBSDIR . '/Boards.subs.php');
```
emanuele45: I wonder when we will not have to deal with requires anymore. :laughing: 
/me dreams of autoloading everything. :sleeping_bed:

emanuele45:
@Spuds I guess the error is the usual one, right?

sources/controllers/Notify.controller.php
Spuds: Yeup  ... mysql no longer provides a 5,7 packages for the trusty build.  

I've set things to only use mysq5.6 for the 1.1 branch as I don't want to do the full xenial update that I did for the 2.0 branch.  Also dropped the 5.4 testing since who cares about that.

PR #3367, opened 2019-11-05
[WIP] First round of miving the management of temporary attachments to its own class

emanuele45:
It's just kind of masking the complexity, though... well it makes code a bit more easy (maybe).
The idea is similar to what I did in the past with another session variable: a sort of wrapper around $_SESSION['temp_attachments'] with the bonus of moving some methods to the class allowing for some work once spread around to be now in a single place.

Still requires some work. :stuck_out_tongue: (And a lot of documentation blocks lol)
Up for comments!

### DIFF
--- a/sources/ElkArte/AdminController/ManageAttachments.php
+++ b/sources/ElkArte/AdminController/ManageAttachments.php
@@ -344,9 +344,6 @@ class ManageAttachments extends AbstractController
 			// Are attachments enabled?
 			array('select', 'attachmentEnable', array($txt['attachmentEnable_deactivate'], $txt['attachmentEnable_enable_all'], $txt['attachmentEnable_disable_new'])),
 			'',
-			// Extension checks etc.
-			array('check', 'attachmentRecodeLineEndings'),
-			'',
 			// Directory and size limits.
 			array('select', 'automanage_attachments', array(0 => $txt['attachments_normal'], 1 => $txt['attachments_auto_space'], 2 => $txt['attachments_auto_years'], 3 => $txt['attachments_auto_months'], 4 => $txt['attachments_auto_16'])),
 			array('check', 'use_subdirectories_for_attachments', 'subtext' => $txt['use_subdirectories_for_attachments_note']),
--- a/sources/ElkArte/AttachmentsDirectory.php
+++ b/sources/ElkArte/AttachmentsDirectory.php
@@ -177,6 +177,7 @@ class AttachmentsDirectory
 				$expected_dirs++;
 			}
 		}
+
 		return $expected_dirs;
 	}
 
@@ -308,13 +309,14 @@ class AttachmentsDirectory
 	/**
 	 * The current attachments path:
 	 *
-	 * What it does: @todo not really true at the moment
+	 * What it does: @return string
+	 *
+	 * @todo not really true at the moment
 	 *  - BOARDDIR . '/attachments', if nothing is set yet.
 	 *  - if the forum is using multiple attachments directories,
 	 *    then the current path is stored as unserialize($modSettings['attachmentUploadDir'])[$modSettings['currentAttachmentUploadDir']]
 	 *  - otherwise, the current path is $modSettings['attachmentUploadDir'].
 	 *
-	 * @return string
 	 */
 	public function getCurrent()
 	{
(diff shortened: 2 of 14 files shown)


Spuds:
I'm still taking it all in !

emanuele45:
@Spuds while breaking everything I can break, I found this:
https://github.com/elkarte/Elkarte/blob/f3d4ce57ce9d04accaaab8b709da3eaafe510229/sources/subs/Emailpost.subs.php#L883
https://github.com/elkarte/Elkarte/blob/f3d4ce57ce9d04accaaab8b709da3eaafe510229/sources/subs/Emailpost.subs.php#L913

they are just few lines apart, but have a different aspect.
I don't have a lot of knowledge of the PbE stuff, by the look of it and thinking it may be working under kind of CLI/cron, I would tend to think the first seems the correct one (because allowedTo may be fail unless the proper user is set up... that could be an option), but I'm not sure at all... what do you think?

Spuds:
I also feel the first one is the correct syntax for the reasons you stated.   I added the attachment support towards the end of working on PBE and TBH its not been tested a lot nor probably used much in real world forums, but its a goodie for sure.

Since you are working on the attachment code, consider just integrating in the attachment resize code, I feel that should be part of the core these days (given the stupid size of photo images these days)

emanuele45:
Spuds wrote on 06/11/2019 22:31:
>
> I also feel the first one is the correct syntax for the reasons you 
> stated. I added the attachment support towards the end of working on 
> PBE and TBH its not been tested a lot nor probably used much in real 
> world forums, but its a goodie for sure.
>

/me knows the feeling :D


> Since you are working on the attachment code, consider just 
> integrating in the attachment resize code, I feel that should be part 
> of the core these days (given the stupid size of photo images these days)
>

Is it that one?
https://github.com/Spuds/Elk_Resize_Attachment_Images

Spuds:
Thats the one ... There is a 1.1 bug for which I posted a potential fix (on the site).  The addon ran when a file was added to the attachment area, and then again when the post was submitted.  

That double action seemed to cause a filestat error *sometimes* ... ensuring it only runs once appears to have solved that.  The fix I posted on the site has it only run during post figuring thats the safest option but of course no longer running on the ulattach action can lead to late failures in the upload which kind of sucks.

The other annoyance is that the addon sets the allowable form size to system max values, otherwise the form would reject the file before we could work on it.  Our attachment form/js only has one size and it really needs 2 ... one for images that we can resize/shrink and one for the rest so if someone tries to upload a big pdf it will be rejected on attach vs on post.

emanuele45:
Since I started heavily messing with attachments I started wondering: do we really need a whole new entry for the thumbnails?

Any attachment that has a thumbnail has two entries in the db, one for the attachment and one for the thumbnail.
From what I understand the main advantage of this solution is mostly to have a fast way to calculate the size and number of files in the directory without having to resolve to probing the file system, and it should allow us to store the attachment in a directory and the thumbnail in another (in case for example we are close to the limit and the image fits, while the thumbnail goes above said limit).
Apart from that, it forces a LEFT JOIN of ```{db_prefix}attachments``` when we want to show the thumbnail in the page.

So I was thinking it:
1) drop the "requirement" of having the second entry for thumbnails
2) move the size to a new column in the attachments table
3) adapt the queries.

Probably not in this PR, though. :smile:

emanuele45:
@Spuds related instead to the image resize: have you ever experimented with client-side resizing?
I was thinking that if we can resize the image client-size, we can avoid changing the server-side logic entirely, so: if you have js enabled it resizes the image client side (saving 1) server load, 2) bandwidth, 3) work for us), if you have js disabled (mostly me alone) it simply refuses the image.

TBH I never tried it, what do you think?

Spuds:
I do remember looking into this using html canvas and its image manipulation functions, but it was several years ago so I'm sure the issues I found are probably no longer relevant.  Its certainly something to revisit to see how its performance and browser support is (especially via smart phones which are the primary suppliers of those large images)

Spuds:
I did some more reading and we will need a library to do this to avoid going backwards in terms of the down sampling quality.    The built in functions in canvas do a rather poor job in terms of the down sampling quality, we need to stay with a Lanczos based filter and a touch of unsharp mask to keep the quality decent.   There are some old browser support issues as well, mainly with allowed canvas size and the ability to create / copy the blobs.  Seems like the Pica library would be a good choice, there are others but what I've found are either massive or use other suspect re-sampling algos.

We may also want to implement the rotation correction code client side (will require a JS exif library), otherwise the displayed icons in the D&D area will be sideways or we simply still correct them server side and leave the interface thumbnail wrong.

I can take a wack at this if wanted, or you can.  Either way while its being investigated we can decouple this PR and the resize stuff so its a step at a time.

Spuds:
@emanuele45  Oops ... I meant to open a PR on your branch 😇

emanuele45:
ehm... okay, if I got it right the commits were pushed directly into the PR, I guess it's all good! :D 
TBH I'm no longer sure what was missing here... :sob:

Spuds:
Yup .. I intended to push to my repo but instead pushed to yours, which Github accepted probably because it was an open PR in the ElkArte repo *shrugs* ... oh well more coming soon since I can bwahahahaha 😈

emanuele45:
I knew I forgot something! :laughing: 
https://github.com/elkarte/Elkarte/compare/development...emanuele45:more_reorg?expand=1
Now I just have to remember what I missed on that part. :man_facepalming:

Spuds:
I knew there was more to come!

PR #3370, opened 2019-11-20
Menus Update Part XXVII

Spuds:
This is a continuation of  https://github.com/Spuds/Elkarte/tree/createmenusubs and https://github.com/elkarte/Elkarte/pull/3050 in an ongoing saga to tame this monster.    

Will need some continued refinement as we find the bugs and other rough edges and check things like the hook calling, compatibility and even using the new methods directly in all places.   But right now it seems to be working and I'm tired of looking at it :P

Also as always there were a few non related bugs I uncovered while working on this, so those are in the PR.  

Lastly there were some more bugs in the testing code that this PR also fixes (and why the effected file count is so high), I think that using the phpunit convert to to exception was/is quietly hiding bugs (and not running the assertions either) and I think I got all of those, but you know the old test bed can be a major source of considerable pain in arse.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,3 +69,6 @@ after_script:
 # Output screen shots, error logs etc
 after_failure:
   - /var/www/tests/travis-ci/setup-failure.sh $DB $TRAVIS_PHP_VERSION $WEBTESTS $COVERAGE
+
+after_success:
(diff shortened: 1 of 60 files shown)


Spuds:
https://github.com/elkarte/Elkarte/pull/3379 pulled the bug fixes and test updates out of this, which I knew would break this PR but had to move on.  This PR has commits dating back 3+ years so is just to difficult to maintain any-longer, needs to be redone if the updates to that menu subs beast is wanted.

PR #3371, opened 2019-11-22
Lets see what CodeCov can do

Spuds:
been having some trouble with scrutinizer not finding the coverage report (TT is open with them).  In the meantime figured I try other options.  First attempted codacy but .... now lets try codecov to see what it does

### DIFF
--- a/.codecov.yaml
+++ b/.codecov.yaml
@@ -1,0 +1,7 @@
+codecov:
+  require_ci_to_pass: no
+
+coverage:
+  range: 50..100
+  round: down
+  precision: 2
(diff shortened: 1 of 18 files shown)


Spuds:
Scrutinizer fixed the issue with the coverage report, but leaving codecov in the mix seems like a good thing, it was easy to setup and the reports are decent.   

Also just to note, the coverage % is now up to 21% take that coverage bots!  OK so there is still a long way to go but at least its moving again 😄

emanuele45:
YAY! We have charts! :D 
/me likes colors

PR #3377, opened 2019-12-21
Some more 1.1.7 updates and fixes

Spuds:
Why are you reading this, does anyone read this? Just see the commits instead.


Spuds:
Well that is a gigantic pain in the ass
What the F allows that to, not only run fine, but be required to even run , on my local only to have travis crap its pants .  I guess I'll deal with the tomorrow, maybe, right now I'm just pissed.

Spuds:
Thats enough for this PR, so if you like it and don't see any errors .....

emanuele45:
Maybe do you want to add also #3385 or should I not be **that** lazy and push them myself?

Spuds:
No problem, I'll add them in ... 

I also have a pass of the one click unsubscribe about ready for review, and in doing that found a couple of more minor notify/mention bugs that I fixed i(n that branch though)

Spuds:
Why you clever bastard !!! Now I see why you wanted me to add them LOL!

emanuele45:
That was unintentional!! :laughing:

PR #3378, opened 2019-12-22
Push me, go ahead

Spuds:
This makes some changes to the desktop aka push notifications.

- Moves to ```pushjs``` which is the follow on "library" to the desktop notification js that we currently use.  Much is the same but more refined and better browser support.  Still a bit broken in places but better.

- Uses a service worker which allows the notifications to come through on android, well at least android chrome on my phone.

- Now asks for browser notification permission after profile save when they enable notifications vs upfront or when the first notification appears.

- Changed the ajax call to hopefully be more persistent.  Using ```timeout``` and ```.always``` should ensure that ```setTimeout()``` is continually set.


emanuele45:
I think I completely missed some stuff... :-\

PR #3387, opened 2020-01-01
Some more milk with the cookies

Spuds:
This fixes a couple of minor bugs with setting incorrect cookies when a password is changed.  Introduced probably back in 1.0 😭 

This also increases the salt length from 4 to ~~8~~ 16 which provides a vast increase in probable values.  It does not force it to happen but it will use it for new members and do updates on existing members during logout and password changes.

In 2.0 all of this cookie code should be centralized in a class, for setting/getting/clearing


Vekseid:
I'd recommend going to 16 characters for good measure. No reason to be stingy. : )

Spuds:
I don't know why I stopped at 8, probably I was like whoa thats a big number 62^8 but no reason to stop there ... I was planning on going back to this to take a second look anyway.  In 2.0 I'll move the cookie to bcrypt so it really takes up some cpu time, no sense in using any fast hash functions for this.

Vekseid:
The cookie doesn't need (and shouldn't use) bcrypt. Way too intensive for session logging. For 2.0, if we want something ultra secure, we should have a token table for login sessions.

PR #3388, opened 2020-01-01
Smtp 

Spuds:
This fixes an old todo in the code where the HELO EHLO is supposed to be followed by the FQDN of the client, the code instead sent the recipient.   Probably worked fine on a same server but depends on what rules they put in place in postfix for example.    

I did not cache or save the result in the PR, seems like that value would not change so that may be worth considering since to validate FQDN requires some network calls which could? be slow? What do you think?


Vekseid:
I fear this can create a situation where this might get stuck on a bad value, that might also get rejected. I'd rather have an automatically generated setting that I can override, rather than something like this.

Spuds:
*nods* ... I had a few concerns as well, as it certainly could create an incorrect value.  TBH I'm surprised it worked at all prior to this ... it probably had issues if you were using an external SMTP service, or any SMTP receiver that was setup correctly.  

We could call the function to seed an initial 'client' value in the ACP which of course can then be changed and saved.  Just need to add a new input field and txt string.

PR #3389, opened 2020-01-06
Unsubscribe improvements

Spuds:
This extends  the current unsubscribe links from topics/boards to all our various mentions (quoted, liked, mentioned, etc) to allow easy unsubscribe.  It also adds additional information to those mentions, namely the subject in which the event occurred, thats a bit nicer for the email.

Adds the static frequency value to the mention controllers, currently all frequencies were shown for all methods but that was not really correct (e.g. bounced is only notify not email).  With this the method allowed frequencies is shown in the ACP.

links are all now a one button affair with some protections to avoid spamming and general mischief.  I'm not super happy with some things but in general the token is a blowfish crypt with user and extra site salt, this makes a brute force a time consuming process and all you would get is the ability to unsubscribe someone.

Adds a buh-bye thanks for no longer using our quota confirmation page with just what you have been unsubscribed from.

### DIFF
--- a/sources/SiteDispatcher.class.php
+++ b/sources/SiteDispatcher.class.php
@@ -102,7 +102,6 @@ class Site_Dispatcher
 		'moderate' => array('ModerationCenter_Controller', 'action_index'),
 		'movetopic' => array('MoveTopic_Controller', 'action_movetopic'),
 		'movetopic2' => array('MoveTopic_Controller', 'action_movetopic2'),
-		'notify' => array('Notify_Controller', 'action_notify'),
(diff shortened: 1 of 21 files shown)


emanuele45:
Sooo much stuff changed, I'll trust you again on that! :D

Spuds:
I thought the same thing while I made the updates ... darn thats a lot of edits!  I really thought it would be easy since we had unsubscribe in the PBE links, but not for like, quote, mention, rlike.   Oh Well it was kind of easy, just was a lot more edits than I anticipated.

We will need to use 1.1.7 live on the site for a bit since its getting to be a big boy.

PR #3390, opened 2020-01-07
Some more milk with the cookies

Spuds:
Had to close and redo this one as a wrong commit was in this original PR (one from another PR) and in fixing that things went bad quickly!

This fixes a couple of minor bugs with setting incorrect cookies when a password is changed. Introduced probably back in 1.0 😭

This also increases the salt length from 4 ~~to 8~~ to 16 which provides a vast increase in probable values. It ~~does not~~ does force it to happen but should not cause a logout but instead update the salt and cookie on the fly.  It will use it for new members and do updates on existing members during logout and password changes.

In 2.0 all of this cookie code should be centralized in a class, for setting/getting/clearing

### DIFF
--- a/install/Install_Controller.php
+++ b/install/Install_Controller.php
@@ -1067,7 +1067,7 @@ class Install_Controller
 			{
 				require_once(SUBSDIR . '/Auth.subs.php');
 
-				$incontext['member_salt'] = substr(sha1(mt_rand() . microtime()), 0, 16);
+				$incontext['member_salt'] = substr(base64_encode(sha1(mt_rand() . microtime(), true)), 0, 16);
 
 				// Format the username properly.
 				$_POST['username'] = preg_replace('~[\t\n\r\x0B\0\xA0]+~', ' ', $_POST['username']);
--- a/install/Install_Controller.php
+++ b/install/Install_Controller.php
@@ -1067,7 +1067,7 @@ class Install_Controller
 			{
 				require_once(SUBSDIR . '/Auth.subs.php');
 
-				$incontext['member_salt'] = substr(md5(mt_rand()), 0, 4);
+				$incontext['member_salt'] = substr(sha1(mt_rand() . microtime()), 0, 16);
 
 				// Format the username properly.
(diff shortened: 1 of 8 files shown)


install/Install_Controller.php
```diff
@@ -1067,7 +1067,7 @@ private function action_adminAccount()
 			{
 				require_once(SUBSDIR . '/Auth.subs.php');
 
-				$incontext['member_salt'] = substr(md5(mt_rand()), 0, 4);
+				$incontext['member_salt'] = substr(sha1(mt_rand() . microtime()), 0, 16);
```
Vekseid: These return hexadecimal strings, so a better line might be

>  = substr(base64_encode(sha1(mt_rand() . microtime(), true)), 0, 16);

install/upgrade.php
```diff
@@ -869,7 +869,7 @@ function checkLogin()
 					if ($valid_password)
 					{
 						$password = validateLoginPassword($_POST['passwrd'], '', $_POST['user'], true);
-						$password_salt = substr(md5(mt_rand()), 0, 4);
+						$password_salt = substr(sha1(mt_rand() . microtime()), 0, 16);
```
Vekseid: These return hexadecimal strings, so a better line might be

>  = substr(base64_encode(sha1(mt_rand() . microtime(), true)), 0, 16);

Vekseid:
Suggest using base64_encode on the raw output instead of the plain hex output. /pedant.

Spuds:
Fair enough observation ... Should probably fix the strings passed to setLoginCookie as well, I think they are all hexadecimal as well, of course those are also 64 characters long to help protect them.

Vekseid:
https://threatpost.com/exploit-fully-breaks-sha-1/151697/

Suggest switching to sha256 anywhere we're using sha1, just as a matter of principle.

Spuds:
Interesting read.  I don't think we use sha1 in many places as most everything was moved to sha256 long ago.  The remaining ones are only to generate a salt AFAIK, although OpenID uses sha1 it but thats another bag.  All crypto level things use bcrypt and random_bytes.

2.0 I'm considering moving the existing sha256 to sha3-256 as its a bit more capable just for good measure (and since its available since 7,1)

PR #3391, opened 2020-01-07
Menus Update

Spuds:
This is a continuation of .... 🥁 

- https://github.com/elkarte/Elkarte/pull/3370 
- https://github.com/Spuds/Elkarte/tree/createmenusubs
- #3050

Since the work on this goes back almost 3 1/2 years to 1.1 beta time, I did some creative cherry picking and commit combining to get this into something that looks vaguely coherent for 2.0.  The previous PR had bug fixes found while doing this and the PR should be just related to the menu and its collateral

Will need some continued refinement as we find any bugs and other ares like the hook calling, compatibility and using the new methods directly in all places. But right now it is working.


emanuele45:
I think I'm going to trust you on that one, I guess it's wiser than trusting me! :laughing:

Spuds:
Well on a positive note, I did not change the old menu tests at all and they seem to pass, so 🙏 that is a good sign.

PR #3394, opened 2020-01-22
Some more changes to the images generation/handling and some fixes

emanuele45:


emanuele45:
@Spuds I think this one should be _fine_ too. Not sure what I broken, but... whatever. :laughing:

Spuds:
In it goes !

PR #3401, opened 2020-05-31
Extending once again Notifications

emanuele45:
And that should be it this time.
I split up the mentions, the notifications methods, etc. so that it should be quite more easy to use and expand.
The admin can pick the notification methods to allow the members to use and which are the defaults, then each member can decide which type of notification method to pick.
I also split up the email notification into _direct_, daily, weekly so that each can be selected separately, probably this has been overkill, though... whatever, I think it was the easiest way in terms of implementation.

Admin UI:
image

Profile UI:
image

the multi-select is not exactly perfect and on mobile may not be the best option, though in normal conditions seems the easiest approach to me.

This time the code should be configured in a way that is fully extendable, that was the main goal.


emanuele45:
Oh well, I'll merge this one. :innocent: 
If there is bugs... I'll make more!

PR #3402, opened 2020-06-21
Show again the captcha configuration when gd is available

emanuele45:


Vekseid:
...how did this happen? O_o

PR #3405, opened 2020-08-25
Fixes for read-only source files

nwsw:
When the forum files are read-only:

1. The `index.php?action=admin;area=maintain;sa=routine;activity=version` command crashes because the `$tmp_version_info` variable is not set in the `getFileVersions` function in Admin.subs.

2. A custom theme folder cannot be installed because the ManageThemes controller incorrectly looks for `query` parameters when identifying the `$method = 'path'` mechanism. The submission is actually in a `post` request.

This request solves these two problems.


emanuele45:
Nice, that was my doing and I don't even remember what I was trying to achieve. One thing is sure: $tmp_version_info is never set. :facepalm:

nwsw:
>  I don't even remember what I was trying to achieve

This code distinguishes the Subs and Modules folders for recursive scans (when is_writable is true).

PR #3406, opened 2020-08-28
Fix `undefinedtheme` errors in JavaScript

nwsw:
The JavaScriptEscape function actually references the js variable `elk_scripturl` so it is vital that `elk_scripturl` be first in
the `addJavascriptVar` array.

### DIFF
--- a/themes/default/Theme.php
+++ b/themes/default/Theme.php
@@ -1005,11 +1005,11 @@ class Theme extends \Theme
 
 		// Default JS variables for use in every theme
 		$this->addJavascriptVar(array(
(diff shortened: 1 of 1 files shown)


nwsw:
After thinking about this, I realize that this issue is caused by our custom setting for `$scripturl`:

```php
$scripturl = $boardurl . '/';
```
The standard `$scripturl` does not generally trigger this issue, so I am withdrawing this fix.

Note: I'll probably remove the `elk_scripturl` substitution in `addJavascriptVar` from our fork, as it seems pointless.

PR #3416, opened 2020-12-15
PHP 7.4 updates and "other"

Spuds:
This fixes any 7.4 errors that I have encountered, could be more but this is a start.  Some of the 7.4 issues were in fact minor real errors so those were fixed as well

Noticed a bug in the re-encode image function when using imagik, it was not stripping the exif headers and that is where mischievous code could exist, so it was 100% ineffective.

Fixed a few errors reported on the site or in the issues log

Made updates to PBE based on reported errors as well as things found while giving it a good thrashing.   There were some new email content types to deal with and the plain text email link issues.  That fix was a bit eh? but it does what is needed.

Make updates to the build environment which was failing (mostly due to xcache) but also reduced the number of environments.  I'm not sure what is going on with Travis but things sit in the queue for hours and then would only run 4 concurrent jobs, so a 5 job run could take 4 hours to complete, so much for CI ... not sure if this is just the .org to .com change or due to other reasons.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,34 +1,24 @@
-sudo: required
 language: php
 dist: trusty
+os: linix
 
-matrix:
+jobs:
   include:
-    - php: 5.6
-      env:
-        - DB=mysql-5.6
-        - WEBTESTS=false
-        - COVERAGE=true
     - php: 7.3
       env:
         - DB=mysql-5.6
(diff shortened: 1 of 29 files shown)


Spuds:
i-dont-always-merge-pull-requests-but-when-i-do-it-is-usually-my-own-pull-request

PR #3423, opened 2020-12-23
Fix GROUP BY issue on pgsql

tinoest:
Fixes: 

Type of error: database
Database Error: ERROR: column "lmr.id_msg" must appear in the GROUP BY clause or be used in an aggregate function

/var/www/html/PGElkarte/SSI.php
Line: 511


Spuds:
Thanks for the catch and fix!

Issue #2986, opened 2017-09-03
SVG the smiley message icons?


Vekseid:
So I noticed these got 'converted' in part, minus the selectors and thread display. I thought 'Hey I can do that quick'

Then I noticed the smileys have been integrated in some sort of unholy hybrid spawned by @Spuds 

Is there some reason an SVG version was not found or made for these?

emanuele45:
Smiley or message icons? Asking just to be sure we are talking about the same thing. :wink: 
I feel it's message icons.

Vekseid:
The message icons.

Five of them are still .png, all smileys.

There's a rather ugly bit of code that enables the quick-icon select, but everything else is pretty straightforward if we can get those five made into .svgs

Spuds:
Sadly I only converted the ones that were sprited in my unholy attempt.  In any case we still need to support png message icons for 1.1 but those last 5 standard ones could be svg'ed should suitable ones be found.

emanuele45:
In general, I would say that support **images** is the way to go: many may have images they want to use as message icons and don't can/know/want to convert them to svg. As ugly as it may be.

Vekseid:
How many Elkarte users are changing the message icons? Hell many themes don't even do that.

emanuele45:
I don't have current data.
I know in the "past" it was rather common.
I know we have an admin interface to add new message icons (and I even wrote a mod for SMF to extend it and make it more flexible) and asks for png.

For 1.1 I would not remove the option to let admins add new message icons.
For 2.0 I guess we can talk about it (worst case we can extract the code and create an addon).

Vekseid:
With the svg version you can add any svg icon you want to the list. That doesn't go away.

Vekseid:
@emanuele45 @Spuds Are you familiar with this: https://github.com/EmojiTwo/emojitwo

Thousands of SVG smilies. A lot of popular software uses version 3 of these (under a more restricted license).

Spuds:
I noticed icon moon has the needed SVG smile's so I've added those to the mix.  Its still an unholy terror but at least in the message listing the standard icons will now all be SVG.  There are still areas that use the PNG version of those as well but not sure we can do anything about those areas until 2.0

Spuds:
With that PR, I'm going to call this good for now ... really only one area to address.

PR #3424, opened 2020-12-24
Msg Icons

Spuds:
This is to partially address #2986

1.  adds SVG version of "smiles" to the message icons
2. uses the SVG version on topic listings like recent and standard message index.  It will use these as long as the admin has not enabled custom message icons.  Once custom icons are set it will revert to the old .PNG way
3. does not work (nor did it) for SVG on the display template since that icon is JS linked to allow quick selection and it is simply to much work to fix it for a single icon display
4.  while doing this I noticed that the  MessageTopicIcons.class.php was wired to assume the custom icons were those defined in the topic_listing, however there are some potential problems, such as a custom icon was allowed on a board and then removed from that board, current behavior was that it will still render ... other similar issues that resulted in no icons (missing image).  As such I made several changes to that class as follows:

- it loads the custom icons for the given board so it will only render the allowed ones and set the default for wrong/missing/changed ones.  This is a cached call so not much additional overhead.
- you can still pass it a custom array to maintain old behavior, although the array index has changed, sooooo
- added two get classes, one for url and one for name.  The name is really for svg icons, so again say you enabled custom icons, added some, they were used, then you went back to standard behavior.  Well it would have shown an empty icon but it now shows the default one.
- not really about this class but I moved the fetch board message icons to the message icon subs where it seemed like it should live (vs inside editor.subs)
- updated the various calls to use the getters vs magic __get

OK thats enough for now @emanuele45 if you could take a look I'd appreciate it!


emanuele45:
Looks good to me

PR #3427, opened 2020-12-28
Emoji

emanuele45:
@Spuds I picked your addon and stuffed in 2.0. :angel: 
Let me know what you think :grin: 

https://www.elkarte.net/community/index.php?topic=4695.0


emanuele45:
I was looking into a UI to pick the emoji, but for the time being I gave up.
I noticed some flag emoji are missing and I also noticed some emoji have odd keys, like the it flag with '1f1ee-1f1f9' :raised_eyebrow: that looks like the range rather than the exact key, not sure if it is supposed to be like this or otherwise how it works. :man_shrugging:

Spuds:
I will look in just a bit ... busy pulling in all the 1.1.7 commits into 2.0 ... have pity on me, its a conflict monster !!!!

emanuele45:
I have the general idea. :laughing: 
If you want to leave it hanging around I can have a look after 1.1.7 is released. :innocent:

Spuds:
I have this working on my local so will try to make a PR to your branch .... Had to fix a couple of things that were either broken my 1.1 version or ??  anyway, also found a small bug with sceditor and will push an update for that as well (had to do with emoticons) Since this is core I'm making a few other adjustments.  Also will re release it under elkarte license since I can!

Spuds:
@emanuele45 to answer another question, I had to look back at when I did this ... to keep the list more contained I removed a bunch of emoji that I felt were extraneous, so I removed things like all the flags and some other things.  The full list is like 3000 items and I culled it to around 1000 so that is why some are missing.

Spuds:
And one last item was that since it uses two emoji sets, but since they are on different "standard emoji levels" they don't have 100 correlation, so the ones I choose were the ones that existed in both sets ... I went through the latest versions of teh two this PM and there are about 200 more that we can add, taking the total up to ~1050 (still excluding flags) so that change is coming also 👯

Spuds:
@emanuele45 pr made to your branch 🕺

emanuele45:
I guess this will not work well with the SCEditor updates, right?
I'll try a rebase in the next days.

Spuds:
Yup it has problems with the new editor.  I've rebased to dev and am working on the fixes.   Going to need to fix mentions as well, I thought it was fine, but not so much 😭

PR #3428, opened 2020-12-29
Query to fetch

emanuele45:
The original idea was to convert the "query" without identifier to "queryFetch", but I re-scoped to drop the unnecessary identifiers still hanging around.
There are still a few that I'm not sure how to get rid of, though... it's getting closer.

### DIFF
--- a/SSI.php
+++ b/SSI.php
@@ -322,7 +322,7 @@ function ssi_queryPosts($query_where = '', $query_where_params = array(), $query
 	$db = database();
 
 	// Find all the posts. Newer ones will have higher IDs.
-	$request = $db->query('substring', '
+	$request = $db->fetchQuery('
 		SELECT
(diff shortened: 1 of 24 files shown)


emanuele45:
@Spuds looks like to fail on fetching elkarte.net?
Maybe firewall or such?

Spuds:
We should update the test to https .... but I'd guess is that Travis uses its own DNS resolver and it did not respect the TTL I set on the old site in prep for the move.  If so it should self clear whenever Travis decides to refresh its resolver.   Or its just another test issue !

Spuds:
@emanuele45 could you resolve the conflicts caused by the1.1.7  🐻 so we can get this merged in 👍

emanuele45:
:crossed_fingers:

Spuds:
Yeah!

PR #3429, opened 2020-12-30
Munge117 into 20

Spuds:
Well here it is ... all of the changes (thus far) from 1.1.7 munged into 2.0

Some changes were not needed
Some changes needed changes
Some changes were difficult to put in place and I did not test them 😇 

With all of the file name changes, {} formatting, refactoring and moar !!! it was a long haul to get this in place.  But it says it will merge which is a miracle!

I would say in the future, given how far 2.0 has grown from 1.1 we really should consider merging into 2.0 at any 1.1.x PR time since its really not an easy thing to do at this point.   Then again after 1.1.7 I don't really want to do much on 1.1.x 😸 and want to get back to 2.0


emanuele45:
> I would say in the future, given how far 2.0 has grown from 1.1 we really should consider merging into 2.0 at any 1.1.x PR time since its really not an easy thing to do at this point. 

IKR :laughing: 

> Then again after 1.1.7 I don't really want to do much on 1.1.x smile_cat and want to get back to 2.0
Yup
Actually, I was starting to consider to move elk.net to 2.0 dev while we are at it...
I may test it locally on the 1st and if nothing goes terribly wrong I can do the move on the 2nd or the 3rd.

Spuds:
Agree ... it will need this PR to run on 7.4 as well.  The old site would not even support 2.0 so that was also part of my reason for asking TE for some assistance. 

I grabbed a couple of minor things out of the site log that I'll toss in the 1.1.7 "final" and keep an eye out for anything else that may come up.  I'l like to wrap that up release in the new year, I can't bring myself to release anything in 2020!

sources/ElkArte/Controller/Xml.php
```diff
@@ -757,12 +756,12 @@ public function action_parserorder()
 		// Failed validation, tough to be you
 		else
 		{
-			if (!empty($validation_session))
+			if ($validation_session !== true)
```
emanuele45, line 759: hmm...
Shouldn't this be 
```
($validation_session === true)
```

sources/ElkArte/Controller/Xml.php
```diff
@@ -856,12 +855,12 @@ public function action_messageiconorder()
...
```
emanuele45, line 858: Same as above

sources/ElkArte/Controller/Xml.php
emanuele45: hmm.... okay, never mind, actually the old one seems pretty broken, because validateSession was returning either `true` or a string, so technically it was always not empty.

sources/ElkArte/EventManager.php
```diff
@@ -235,47 +237,34 @@ protected function _register_events($classes)
...
-	{
-		if ($this->_declared_classes === null)
+		catch (\Exception $e)
 		{
-			$this->_declared_classes = get_declared_classes();
+			$number_params = 0;
 		}
 
-		return $this->_declared_classes;
+		// Php8 will not like passing parameters to a method that takes none
```
emanuele45, line 264: php8 is annoying. :laughing:

sources/ElkArte/Http/BrowserDetector.php
```diff
@@ -472,7 +472,7 @@ public function isPossibleRobot()
 		// Be you robot or human?
 		if (!isset($this->_browsers['possibly_robot']))
 		{
-			if (User::$info->possibly_robot)
+			if (isset(User::$info->possibly_robot))
```
emanuele45, line 475: Just wondering: is this php8 giving errors again?

sources/ElkArte/Controller/Unread.php
```diff
@@ -334,20 +334,7 @@ public function action_unread()
...
-		}
-
-		// All unread replies with temp table
-		if ($context['showing_all_topics'] && $this->_grabber->hasTempTable())
-		{
-			$this->_num_topics = $this->_grabber->numUnreads(false);
-			$type = 'message';
-		}
-		// New posts with or without temp table
-		elseif ($this->_is_topics)
```
emanuele45, line 350: Is the removal of this intended?

sources/ElkArte/Unread.php
```diff
@@ -633,6 +506,6 @@ private function _getUnreadReplies($start, $limit, $include_avatars = false)
 		}
 		$request->free_result();
 
-		return TopicUtil::prepareContext($return, true, ((int) $this->_preview_bodies) + 128);
+		return Topic_Util::prepareContext($return, true, ((int) $this->_preview_bodies) + 128);
```
emanuele45, line 509: This seems to be not correct. I think the class name was changed to TopicUtil in 2.0

sources/ElkArte/Controller/Xml.php
Spuds: yup, it had been broken for a bit.

sources/ElkArte/Http/BrowserDetector.php
Spuds: I don't remember if this was 7.4 / 8 issue, but for sure if things don't 100% pass back a bool then compares break.  If I either saw it while running or in the error log it got "patched"

sources/ElkArte/Controller/Unread.php
Spuds: I'll double check.   It would have been part of https://github.com/elkarte/Elkarte/commit/9eff7c2e21325a3096ef63f19579ef6072727d45 where all the temp tables were removed or git could have gone wild.

sources/subs/Topic.subs.php
```diff
@@ -1708,16 +1701,21 @@ function selectMessages($topic, $start, $items_per_page, $messages = array(), $o
...
+			SELECT 
+				m.id_msg 
+			FROM {db_prefix}messages AS m
 		WHERE m.id_topic = {int:current_topic}' . (empty($messages['before']) ? '' : '
 			AND m.id_msg < {int:msg_before}') . (empty($messages['after']) ? '' : '
 			AND m.id_msg > {int:msg_after}') . (empty($messages['excluded']) ? '' : '
 			AND m.id_msg NOT IN ({array_int:no_split_msgs})') . (empty($messages['included']) ? '' : '
 			AND m.id_msg IN ({array_int:split_msgs})') . (!$only_approved ? '' : '
 			AND approved = {int:is_approved}') . '
 		ORDER BY m.id_msg DESC
```
emanuele45, line 1714: Nitpicking it should be indented by 1 tab I think. :angel:

tests/sources/subs/HTML2Md.class.Basic.php
```diff
@@ -30,7 +30,7 @@ public function setUp()
 			array(
 				'Named links',
 				'<a href="http://www.elkarte.net/" class="bbc_link" target="_blank">ElkArte</a>',
-				'[ElkArte](http://www.elkarte.net/)',
+				'[ElkArte](http://www.elkarte.net/ 🔗)',
```
emanuele45, line 33: :raised_eyebrow:

sources/ElkArte/Unread.php
Spuds: Oops you are correct!  I caught some of those during the munge ... thats what made this so much fun LOL

emanuele45: I can imagine! :rofl:

sources/ElkArte/Controller/Unread.php
emanuele45: It happened so long ago I completely forgot! :facepalm:

sources/ElkArte/Http/BrowserDetector.php
emanuele45: Good to know!

sources/subs/Topic.subs.php
Spuds: Anything to help make sub selects a bit easier to read

tests/sources/subs/HTML2Md.class.Basic.php
Spuds: This is a trick ... and maybe not the best, but here is the long story.

When we convert links to markdown for email output the final ')' of the ```[]()``` syntax can cause problems with email clients .. so if your link ended in a ) ,which is a valid link character, then when converted to MD the email client will see it as a )) and  boom bad link.  That is just one example of rouge links.

Adding in that utf8 character forces our outbound email to also output a base64 version (since there is a non ASCII character) in addition to the plain text version.  The base64 version is much less susceptible to getting munged up in an email client (specifically wrong text wrapping) and email clients will prefer to show that over the plain text.  If you look at the plain text you would just see a &#1234 thingy at the end but the link would still work.  It only does this trick when needed.  We could just add a space and the link would then work, as long as it did not get wrapped by the client, but we end up loosing the base64 version in the email

I also fixed the wrong calculation of line wrapping so at least in the plain text version we don't break the link, however some email clients do this on their own.

sources/ElkArte/EventManager.php
Spuds: Its VERY particular about parameters now, number passed, ordering,  optional ones need to be at the end, etc.

Spuds: Hummm...looks like a git merge/munge error ... Guess it thinks public function register() is not important anymore DOH!

Spuds:
I'm getting close to having these tests pass

I fixed a several test issues that were not  "real" ElkArte issues.  

Did find an interesting one where our sub-selects with LIMIT statements like .```... LIMIT 1, 30 ) AS bla```` would fail on Postgre due to the regex not expecting anything after a limit statement,  so that regex needs to change.

Updated the matrix to add PHP 7.4 testing and fixed serval other issues that were real but likely also due the hot wiring we do to run tests,

Spuds:
@emanuele45  Well that was easy, only took two days to get the friggin tests to pass LOL ... Some of it was ElkArte bugs, some were testcases that needed to be updated/improved,  Some was Travis being Travis (where the tests passed on my repo but then failed here)

emanuele45:
:muscle: 

Time to merge?

Spuds:
In it goes !!!

tests/sources/subs/HTML2Md.class.Basic.php
emanuele45: Oooohhh... cool!
Thanks for the explanation!

Issue #3248, opened 2018-10-28
Replace all the URLs with getUrl


emanuele45:
Any place $scripturl is used should be converted to getUrl.

PR #3435, opened 2021-01-12
More fun with getURL

Spuds:
Sure to break things, and still a lot to go, for another day and another PR.   All part of #3248


emanuele45:
> Sure to break things
Have we ever cared about not breaking stuff? :metal:

PR #3436, opened 2021-01-13
Post Updates

Spuds:
Just another round of breaking up super long post methods to just very long post methods

Included some other items from static analysis and trying to fix docbloc throws methods wich storm seems to like to add to methods even though they don't throw them (cascade effect) which just makes things harder to traverse IMO

### DIFF
--- a/sources/ElkArte/Database/QueryInterface.php
+++ b/sources/ElkArte/Database/QueryInterface.php
@@ -55,7 +55,7 @@ interface QueryInterface
 	 * @param string $db_string
 	 * @param mixed[]|false $db_values = array()
 	 *
-	 * @return bool|Result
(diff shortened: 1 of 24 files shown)


sources/ElkArte/Database/QueryInterface.php
```diff
@@ -66,7 +66,7 @@ public function query($identifier, $db_string, $db_values = array());
 	 * @param string $db_string
 	 * @param mixed[] $db_values = array()
 	 * @param mixed[]|null
-	 * @return bool|Result
+	 * @return bool|\ElkArte\Database\AbstractResult
```
emanuele45, line 69: I wonder if I lost a ResultInterface somewhere or if I left it out on purpose...

emanuele45:
> (that are still to long

The longer the better!

PR #3437, opened 2021-01-13
Attach updates

Spuds:
Some more attachment love

### DIFF
--- a/sources/ElkArte/TemporaryAttachment.php
+++ b/sources/ElkArte/TemporaryAttachment.php
@@ -267,7 +267,7 @@ class TemporaryAttachment extends ValuesContainer
 					}
 				}
 			}
-			catch(\Exception $e)
+			catch (\Exception $e)
 			{
 				$this->setErrors('bad_attachment');
 			}
--- a/sources/ElkArte/TemporaryAttachment.php
+++ b/sources/ElkArte/TemporaryAttachment.php
@@ -36,6 +36,8 @@ class TemporaryAttachment extends ValuesContainer
 	 * Deletes a temporary attachment from the filesystem
 	 *
 	 * @param bool $fatal
+	 * @return bool
+	 * @throws \Exception
 	 */
 	public function remove($fatal = true)
 	{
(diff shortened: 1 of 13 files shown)


sources/ElkArte/TemporaryAttachment.php
```diff
@@ -34,30 +33,43 @@ public function __construct($data = null)
...
+			}
+
+			unlink($this->data['tmp_name']);
+		}
+		catch (\Exception $e)
+		{
+			return false;
+		}
+
+		return true;
```
emanuele45, line 64: Maybe slightly more readable as something like this?

```
if (!$this->fileWritable())
{
	if ($fatal)
	{
		throw new \Exception('attachment_not_found');
	}
	else
	{
		return false;
	}
}

unlink($this->data['tmp_name']);
return true;
```

Unless unlink may throw an exception on failure, not sure.

sources/subs/Attachments.subs.php
```diff
@@ -319,10 +319,11 @@ function createAttachment(&$attachmentOptions)
 	if ($is_image)
 	{
 		$attachmentOptions['mime_type'] = getValidMimeImageType($size[2]);
+
 		// Want to correct for phonetographer photos?
 		if (!empty($modSettings['attachment_autorotate']))
 		{
-			$image->autoRotate();
+			$image->autoRotateImage();
```
emanuele45, line 326: lol
I guess I used autoRotate on the assumption that "Image" was implicit since the object is an image, then I forgot to update the method name in the class. :man_facepalming: 
Oh well, autoRotateImage is fine either.
Thanks for the catch!!

sources/ElkArte/TemporaryAttachment.php
Spuds: It does :(  ``` An E_WARNING level error will be generated on failure.```  so I was just trying to catch both issues with one catch, either no file or permissions or unlink simply 🤮 for some reason.

sources/subs/Attachments.subs.php
Spuds: Easy to rename, not like we are done with attachments just yet ;)

sources/ElkArte/TemporaryAttachment.php
emanuele45: yay I guess. :laughing:

PR #3440, opened 2021-01-19
Update querys from MySQL unique LIMIT offset, start to more universal LIMIT start OFFSET offset

Spuds:
All the db's that we support use that syntax, so this make the core postgre/mysql/mariadb compliant.  I left the postgre translation in place for addons usage but perhaps we should remove that ?


emanuele45:
/me likes this one a lot!

Maybe throw an error if there is any replacement? Worth the hassle?
In any case I would leave it there just in case.

PR #3444, opened 2021-01-31
GHA to replace Travis

Spuds:
This updates our Travis test environment to use GitHub Actions.

On my repo this seems to work fine, the only thing missing are the selenium test (web tests) only because I did not try to move those as I ran out of desire !  You can use this on your own repo by changing the repo name in the test.yml file in .github/workflows

IN all this runs three test matrix jobs, 

- static tests (lint/syntax/headers) 
- mysql/mariadb various db levels vs php matrix tests
- postgres versions vs php versions matrix tests

In my testing, read flailing around trying to get this to run, I did not notice any delay in testing run starts, unlike Travis which in the last several weeks can take hours to start a run.  Add to that the uncertainly of their new pay model etc.

The old travis-ci files are still here, just in a zip file


emanuele45:
Great to see!
I gave it a little try, but I'm not that good with testing stuff.

Spuds:
I've "almost" gotten the web tests to work with selenium so I'll PR what I've found.   The web tests are not fully working, and to be honest I don't think they did on travis either.  I've found issues that simply are not unique to the new GHA testbed.  I thought just getting selenium running and I would be done but 😢

PR #3445, opened 2021-02-01
Db stuff

emanuele45:
I was in a mood of moving stuff around and so I played a bit.
Sending the PR mainly to see if it passes the tests (I can't see them executed on my repo... :man_shrugging: ) but if deemed worth it can be merged.

### DIFF
--- a/install/install_1-1.php
+++ b/install/install_1-1.php
@@ -2636,7 +2636,6 @@ class InstallInstructions_install_1_1
 				array('admin_features', 'cp'),
 				array('last_mod_report_action', '0'),
 				array('pruningOptions', '30,180,180,180,30,7,0'),
-				array('cache_enable', '1'),
 				array('reg_verification', '1'),
 				array('visual_verification_type', '3'),
 				array('visual_verification_num_chars', '6'),
--- a/install/install_1-1.php
+++ b/install/install_1-1.php
@@ -2003,7 +2003,7 @@ class InstallInstructions_install_1_1
 				array('name' => 'mention_type',       'type' => 'varchar', 'size' => 12, 'default' => ''),
 			),
 			array(
-				array('name' => 'mention_member', 'columns' => array('id_member', 'notification_type', 'mention_type'), 'type' => 'unique'),
+				array('name' => 'mention_member', 'columns' => array('id_member', 'mention_type'), 'type' => 'unique'),
 			),
 			array(),
 			'ignore'
@@ -2014,7 +2014,7 @@ class InstallInstructions_install_1_1
 	{
 		return $this->db->insert('ignore',
 			'{db_prefix}notifications_pref',
-			array('id_member' => 'int', 'mention_type' => 'string-12', 'notification_type' => 'string-20'),
+			array('id_member' => 'int', 'mention_type' => 'string-12', 'notification_type' => 'string'),
 			array(
 				array(
 					0,
@@ -2042,7 +2042,7 @@ class InstallInstructions_install_1_1
 					json_encode(['notification'])
 				),
 			),
-			array('id_server')
+			array('id_member', 'mention_type')
 		);
 	}
 
--- a/install/install_1-1.php
+++ b/install/install_1-1.php
@@ -2138,7 +2138,7 @@ class InstallInstructions_install_1_1
 				array('reply_only'),
 				array('read_only'),
 			),
-			array('id_group')
+			array('id_profile')
 		);
 	}
 
--- a/install/upgrade.php
+++ b/install/upgrade.php
@@ -1311,7 +1311,7 @@ function action_databaseChanges()
 			if ($nextFile)
 			{
 				// Only update the version of this if complete.
-				$db->insert('replace',
+				$db->replace(
 					'{db_prefix}settings',
 					array('variable' => 'string', 'value' => 'string'),
 					array('elkVersion', $file[2]),
--- a/install/upgrade_1-1.php
+++ b/install/upgrade_1-1.php
@@ -677,7 +677,7 @@ class UpgradeInstructions_upgrade_1_1
 				'debug_title' => 'Inserting custom fields for gender/location/personal text',
 				'function' => function()
 				{
-					$this->db->insert('replace',
+					$this->db->replace(
 						'{db_prefix}custom_fields',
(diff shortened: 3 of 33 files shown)


Spuds:
MaxPain (like travis) occasionally barfs on the curl fetch to google, so I'll change that to another endpoint.  Likely getting the bot challenge or other.  The other error on 7.4/11.10 postgre may be legit.

Spuds:
Oh to run it on your repo, you will need to change the .github/workflow/tests.yml file ... just change the repo names elkarte/elkarte to yours, and also add in your branches to monitor pull push (not sure if wildcards will work there).  Then probably add that file to the git ignore list.

emanuele45:
Spuds wrote on 2/2/21 12:53 AM:
>
> MaxPain (like travis) occasionally barfs on the curl fetch to google, 
> so I'll change that to another endpoint.
>

Yep, noticed that. Since it is just a GET, maybe github could be used... 
:shrugs:

> Likely getting the bot challenge or other. The other error on 
> 7.4/11.10 postgre may be legit.
>

True... I'm not sure what the problem it, though... I'll try to fix. xD

Spuds:
I'll take a look in just a bit ... I've been updating things to use phpunit 8 so we are no so far behind and there were a lot of depreciations and small changes to help Maximize da pain!   V8 is still supported for like 3 more weeks and also has a min requirement of 7.2 so no more 7.1 tests.

Spuds:
@emanuele45 I just noticed we are getting a ```BLOB/TEXT column 'notification_type' used in key specification without a key length```  notification_type is being used as a key so mysql (at least) wants an index length on that text key ... maybe varchar instead that support 65K I thought, but maybe not with postgre

Spuds:
@emanuele45 The two search errors were due to me forgetting to change the testcase from stuffing ```$_GET['search']``` to ```$req->query['search']```  Some other errors (web and look at mysql) are due to the above mentioned TEXT/BLOB being used in the key which causes MySQL/MariaDB failing on instalation of the DB

Still looking into the PBE thing

Spuds:
I see that ```1) TestDatabase::testTablesPopulated
The number of inserted rows for the table settings doesn't match the expectations (202 -VS- 204)
Failed asserting that 204 matches expected '202'.``
So lets start with getting that DB (and MySql/mariaDB) to install proper before much else.   As a temp work around changing the notification type in install from type ```text``` to ```varchar with size 512``` appears to work for the install and the index

Spuds:
@emanuele45 so playing around with the ```notifications_pref``` table and I learned something.  

I kept getting the data to long for column error with text or varchar(512) when inserting the data which was certainly short enough to fit.  Then I thought maybe it was because it was JSON data and some encoding problem, so changed the column type to JSON and that seemed to work.  However different levels of MySQL and Postgre had issues with using a JSON column in the index. 

Then it dawned on me, 🤦 that stupid error of ```data to long for column``` was about the index not the actual field!  So changing ```notification_type``` back to text or perhaps better varchar 256 and then using ```notification_type(64)``` when creating the index works.  That will use the first 64 characters of the field in the index.  I don't think that field will be longer than 64, and if shorter we could lower that.

emanuele45:
Spuds wrote on 2/3/21 2:31 AM:
>
> I see that |1) TestDatabase::testTablesPopulated The number of 
> inserted rows for the table settings doesn't match the expectations 
> (202 -VS- 204) Failed asserting that 204 matches expected '202'.`` So 
> lets start with getting that DB (and MySql/mariaDB) to install proper 
> before much else. As a temp work around changing the notification type 
> in install from type |text|to|varchar with size 512``` appears to work 
> for the install and the index
>

I *feel* most of these errors are due to the replace method I introduced 
in this branch and the roundabout way to impleent the behaviour. Since 
we are using PSQL 9.5 for testing I assume this is also our lower 
target, so I tried using ON CONFLICT and locally seems to work. On top 
of that, the transactions were not working as expected and this was 
another issue.
Let's see with my latest commit... :fingers_crossed: ! xD

emanuele45:
Oh well... *technically* it passed. :laughing: 
Is it necessary to go and open the test to see that it did not even connect to the server? :angel:

Spuds:
The reason for the duplicate index is the mention test was not updated to use the json array.   https://github.com/Spuds/Elkarte/commit/3f0c957bb579ea00647fde48ef35681137085f5b

I originally thought the notification key was not needed as well, but then that error appeared but its due to the test.  So drop that key (or use the length trick), update the test, have a beer

Spuds:
I'm not sure why the db install step showed passed when in fact it failed to connect as it failed to even install due to the index key length issue.  postgre just skipped the bad tables an mysql simply died. 

Its something to look into to.  It also was the reason the web tests failed, since installing the db failed during that testing, I was to hasty to ignore that as a selenium issue but it was because the db install was failing.

Spuds:
On a positive note, with the mentions and search tests fixed (and a bunch of other busy work), my php unit 8 now passes on the dev branch 😄 https://github.com/Spuds/Elkarte/actions/runs/534623410

Spuds:
@emanuele45 Just in case you were hopeful, postgre did not pass, no connection to the db
```
PostgreSQL server version: 10.15 (Ubuntu 10.15-0ubuntu0.18.04.1)
ERROR:  column "id_group" does not exist
LINE 7:    ('read_only')ON CONFLICT (id_group) DO NOTHING
+ sudo rm -rf /install
```

Spuds:
In my pending test update PR I've added the db install fail detection so in the future those errors should be caught

emanuele45:
Spuds wrote on 2/3/21 9:31 PM:
>
> @emanuele45 Just in case you were hopeful, postgre did not pass, no connection to the db
> PostgreSQL server version: 10.15 (Ubuntu 10.15-0ubuntu0.18.04.1)
> ERROR:  column "id_group" does not exist
> LINE 7:    ('read_only')ON CONFLICT (id_group) DO NOTHING
> \+ sudo rm -rf /install

You broke the magic. :laughing:

Spuds:
My prediction for PR's not passing tests .... Pain, Maximum Pain!  Mr. T-bot

So is this good to go now?   I want to PR in the updated tests with phpunit 8 and other fixes

emanuele45:
Spuds wrote on 2/3/21 10:48 PM:
>
> In my pending test update PR I've added the db install fail detection 
> so in the future those errors should be caught
>
Cool!

> My prediction for PR's not passing tests .... Pain, Maximum Pain! Mr. 
> T-bot
>
> So is this good to go now? I want to PR in the updated tests with 
> phpunit 8 and other fixes
>

Aaaand that one seems all green now! :D
So, up to you!

Spuds:
This has some nice changes!

PR #3446, opened 2021-02-02
GH Actions PartII

Spuds:
This adds back the selenium test cases which *seem* to work
Fixes an error with nginx/php-fpm config from first commit which was causing an access error
Tried moz to replace google in fetch tests to avoid the occasional bot or whatever error
Cleaned up (use statements) the testcases
Now with even more maxpain!
Also added the workflow directory to gitignore so we can have local GHA

### DIFF
--- a/.github/bootstrap.php
+++ b/.github/bootstrap.php
@@ -13,6 +13,13 @@
  *
  */
 
+use BBC\ParserWrapper;
+use ElkArte\Cache\Cache;
+use ElkArte\Debug;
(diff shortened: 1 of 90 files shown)


Spuds:
Thats a strange error, the missing file for web testing IS part of the PR?  Maybe a GHA artifact

Spuds:
Since the other tests pass, and the webtest failed since its missing the file thats in this PR, I'm going to merge this in and 👀

PR #3447, opened 2021-02-02
Lets see if the right repo helps with webtests

Spuds:
Yeah


Spuds:
So part of the errors here are from the notification_type not allowing mysql to install .... working on this I update things to use php unit 8 so closing this one and will reopen a new one with moar changes

PR #3452, opened 2021-02-07
Reorganized methods to load languages and essential theme data

emanuele45:
Okay, the commit is a bit crazy, though... it's actually a monolith, there is no point in splitting it. :man_shrugging:

### DIFF
--- a/.github/bootstrap.php
+++ b/.github/bootstrap.php
@@ -104,6 +104,6 @@ updateSettings(array(
 removeSettings('mentions_member_check');
 
 // Basic language is good to have for functional tests
-theme()->getTemplates()->loadLanguageFile('index+Errors');
+\ElkArte\Themes\ThemeLoader::loadLanguageFile('index+Errors');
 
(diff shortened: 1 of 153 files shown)


emanuele45:
Forgot to mention: the side effect of this is to actually fix the display of generic icons for attachments that are not images, before that loadEssentialThemeData was failing because the theme was not loaded (since technically the download of attachments does not need the template (partially wrong, because in case of these placeholder images we actually need to know the paths), see needTheme in Attachment controller) and it was not possible to locate the file.

Spuds (approved):
It looks good to me ... I think splitting that out is the right thing to do.  Did not see any problems although that search replace left us with one use statements to add back as some point 😄 but there are automated tools for that.

Spuds:
I like it 👍

emanuele45:
> although that search replace left us with one use statements to add back as some point smile but there are automated tools for that.

I am lazy, you know. :laughing:

PR #3459, opened 2021-02-17
Search updates

Spuds:
This makes a few more updates to the sphinx search. Notably it adds a new API method to check if the API can use the extended search query build or the standard.  The way it was, the standard would always run and then be ignored by the sphinx API.  This addresses that and simply streamlines the path.  I think there is work to do in the highlighter function but that is for another time.

Next it removes the small and medium custom search index options leaving only the "large".  The others were really only a full  option for MySQL anyway and the usefulness of those is not even debatable, they are atrocious.

Next up It changes the way the custom database index is created.  The original way was using crypt(word), then using the first 2,4 or 5 "alpha/numeric usable" characters of that 32? character result.  Then multiplying that it by a power of the i'th chr() of the search (i'th being 2,4,5) word, I :poop: you not, that is about what it did.  The idea here is to avoid word collisions in the index but its painfully slow and has a poor distribution.

I changed crypt to a FNV1A32 hash (8 hex result) which is really designed for a db index, fast, very good distribution and uniqueness.   There was a article on running this against s 220K dictionary and 90000 number list and having 4 "collisions" where two non = words hashed out the same a.k.a NBD.  I rather doubt the old method was ever benchmarked for this, but crypt is a security algo not a speed/distribution/etc. algo.  Of course  FNV1A32 was likely not available when that was installed.  This will require a rebuild of any custom search index but its worth it.  To note this algo should be available since 5.7, but if we have a problem its an easy function add if needed.

Lastly it makes some changes so searching for 5.7 is possible without finding 57 as well, its a better overall tradeoff


emanuele45:
When you are ready merge it! :P

Spuds:
I think I'm done with it again ... you know until next time.  On a positive note it builds the custom index much faster, still slow as heck, but faster.  Also fixed an old error that left the index a bit bigger than it could be

PR #3460, opened 2021-02-17
Removed initial default argument on insert

tinoest:
It doesn't have any impact or effect when '' is used, as that evaluates to INSERT on the line starting $queryTitle

### DIFF
--- a/sources/database/Db-mysql.class.php
+++ b/sources/database/Db-mysql.class.php
@@ -708,7 +708,7 @@ class Database_MySQL extends Database_Abstract
 	 * @param mysqli|null $connection = null
 	 * @throws Elk_Exception
 	 */
-	public function insert($method = 'replace', $table, $columns, $data, $keys, $disable_trans = false, $connection = null)
+	public function insert($method, $table, $columns, $data, $keys, $disable_trans = false, $connection = null)
 	{
 		global $db_prefix;
 
--- a/sources/database/Db-postgresql.class.php
+++ b/sources/database/Db-postgresql.class.php
@@ -551,7 +551,7 @@ class Database_PostgreSQL extends Database_Abstract
 	 * @param resource|null $connection = null
 	 * @throws Elk_Exception
 	 */
-	public function insert($method = 'replace', $table, $columns, $data, $keys, $disable_trans = false, $connection = null)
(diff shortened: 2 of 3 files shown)


tinoest:
This probably needs further testing, but I couldn't see any impact. I'm unsure why it was written like that historically.

Spuds:
@emanuele45 do you recall why it was like that?  Its the cause of some error log entries (server not elkarte)

tinoest:
```
PHP Deprecated:  Required parameter $table follows optional parameter $method in /var/www/html/elkarte/sources/database/Db-mysql.class.php on line 711
PHP Deprecated:  Required parameter $columns follows optional parameter $method in /var/www/html/elkarte/sources/database/Db-mysql.class.php on line 711
PHP Deprecated:  Required parameter $data follows optional parameter $method in /var/www/html/elkarte/sources/database/Db-mysql.class.php on line 711
PHP Deprecated:  Required parameter $keys follows optional parameter $method in /var/www/html/elkarte/sources/database/Db-mysql.class.php on line 711
```

Those are the error's if it helps at all.

Spuds:
And I think that will actually fail on php8, its a lot more strict about parameters

tinoest:
> And I think that will actually fail on php8, its a lot more strict about parameters

It's actually raised on php8.0 

https://php.watch/versions/8.0/deprecate-required-param-after-optional

I hadn't noticed my development box was running PHP 8.0, when testing. I thought I had reverted back to PHP 7.4

emanuele45:
Spuds wrote on 2/17/21 9:00 PM:
>
> @emanuele45 <https://github.com/emanuele45> do you recall why it was 
> like that? Its the cause of some error log entries (server not elkarte)
>

TBH I never understood why it was done that way. The method is always 
specified, so no matter what that default is overridden. *shrugs*

IIRC in 2.0 is already gone last time I played with the db stuff.

emanuele45:
Martyn Brown wrote on 2/17/21 6:52 PM:
>
> This probably needs further testing, but I couldn't see any impact. 
> I'm unsure why it was written like that historically.
>

AFAICT it was done like this on first commit without explanations (by 
Grudge).
I guess it was to indicate a sort of default, but since the argument has 
to always be specified it has always been pretty useless, but remained 
just because PHP didn't complain.

PR #3461, opened 2021-02-18
Page index problems

Spuds:
Sooooo when Semantic links are enabled, getURL will properly url encode the result.  However once that is done it can/will cause issues with sprintf.  In this case its the sprintf(s) used in an ```createpageindex``` function.  There are some workarounds but they all appeared to be fragile or clunky, at least what I could come up with.

As such I backed out the ```createpageindex``` that were using ```geturl``` back to what they were (I'd guess were were 1/2 done) , but to help I  did add in a ```{scripturl}``` substitution so no global's were re-introduced.

After that I decided to do a preliminary class for ```constructpagelinks``` to break that "function" up a bit in hopes someday we can do some further enhancements.  Its really just a set of before, continue, current, continue, after set of methods but easier to read than the original jumbo function.

@emanuele45 what are your thoughts about this?  And no I have not (yet) add the docblocks !


emanuele45:
There is sso much stuff going on in constructPageIndex I'm not surprised it broken with getUrl.
I'm not entirely sure how it works right now, maybe an "easy way out" could be to sprintf only the `';start=%1$d'` bit and append the url later, that way it should not break url_encoded stuff.
Or, instead of sprintf pass to the class the array of params of the URL, append the "start" to the array and after all that call getUrl?

:+1: on the class, it may make it easier to read... maybe! :laughing:

Spuds:
As I unfolded all those crazy long 200+ character lines the class kept getting longer!  At least it somewhat readable now.  I'm going to merge this in as is.  We can review options for create page, maybe create a new new method that takes the array like ```getURL``` and then does what it has to do, finally calling ```getURL``` internal to ```CreatePageIndex```  ... I think that would be cleanest ATM.  Right now we can at least see what we are dealing with.

PR #3465, opened 2021-02-21
Header updates

Spuds:
This moves ```header()``` calls to a class that simply accumulates them and then outputs them when requested.  Sending in a duplicate header replaces an existing ones.  Moved most of redirect to this class as well since that was header work with a maintance task.

The goal is to prevent headers already sent errors and allow an easy way to change a header before it ever gets sent.  Its a static method as where we do final output is not consolidated in one area.

Also updated the attachment area, the headers were a bit messy but I think they are all cleaned up now.  I also changed the compressed file output.  It now looks at the mime type and if its ```text/xyz``` it will attempt to compress the file.  This replaces the old file extension sniff.  It also drops the use of the output buffer to compress the file and instead calls gzencode directly on the file and echos that result.  This potentially uses more memory but the old method would not compress larger files due to buffer overflow and might use readfile or might use file get contents or might use fread .. the update simply uses one to keep it simple.

Bonus updates are a couple of bug fixes found while trying to test this.  Also the protection index.php files (which still have standard header calls) were updated to remove that ugly nested dirname stuff

### DIFF
--- a/addons/index.php
+++ b/addons/index.php
@@ -5,10 +5,10 @@
  */
 
 // Look for Settings.php....
-if (file_exists(dirname(dirname(__FILE__)) . '/Settings.php'))
+if (file_exists(dirname(__FILE__, 2) . '/Settings.php'))
 {
 	// Found it!
-	require(dirname(dirname(__FILE__)) . '/Settings.php');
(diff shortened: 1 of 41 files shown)


emanuele45:
Cool!
I started doing something similar a few times and always given up. :laughing:

Issue #3464, opened 2021-02-20
Fulltext search does not search on subject


tinoest:
**Describe the bug**
When using the fulltext search it only matches against the body. 

https://github.com/elkarte/Elkarte/blob/b938078386c2f0b8d1100cf8169fc6618682c920/sources/ElkArte/Search/API/Fulltext.php#L215
https://github.com/elkarte/Elkarte/blob/b938078386c2f0b8d1100cf8169fc6618682c920/sources/ElkArte/Search/API/Fulltext.php#L235

It would be better to match against subject, body and have the index on subject, body. 

Or three different index's, one for subject, one for body and finally one for both.

Spuds:
I assume this is 1.1, but probably true for 1.0 as well, and 2.0 LOL

tinoest:
> I assume this is 1.1, but probably true for 1.0 as well, and 2.0 LOL

It is, I’ll submit the patch for 1.1 tomorrow if you’d like. I don’t have a 2.0 forum but I can write what should work...

Do you want three index’s, or just the one? Three is probably the best one.

Spuds:
Heck yes I'd like that 👍 Thank you!

Spuds:
Closed in #3466

PR #3466, opened 2021-02-21
Change search to look at subject as well as body

tinoest:
This will correct the issue in the first instance, although I would prefer to change the topic only search also. 

What tab space do you use on here? As it all aligns on my side but github show's it as not. 

Also, how do I get the search to follow the same path each time? It seems to cache or similar so I never get the same results repeated and I'm yet to work out where that is.

### DIFF
--- a/sources/admin/ManageSearch.controller.php
+++ b/sources/admin/ManageSearch.controller.php
@@ -12,7 +12,7 @@
  * copyright:	2011 Simple Machines (http://www.simplemachines.org)
  * license:		BSD, See included LICENSE.TXT for terms and conditions.
  *
- * @version 1.1.6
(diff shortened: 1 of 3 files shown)


tinoest:
I can't link issues currently, or set flags but it semi fixes #3464.

Spuds:
The code should all be tab indented, not spaces,  looks like your updates used spaces.  (I guess 1 tab = 4 spaces).  If you remember (I seldom do) but check the file headers and update the version to 1.1.7 when you make an edit.  

Had not seen the changed results but have not done much fulltext testing either.  

Change looks good BTW 👍

Spuds:
Awesome ... thanks again

PR #3467, opened 2021-02-21
Fixes install errors on php 8.0, checks to make sure this->result is an instance of mysqli_result before free'ing

tinoest:
Fixes; https://www.elkarte.net/community/index.php?topic=5822.msg42134#msg42134

### DIFF
--- a/sources/ElkArte/AdminController/ManageSearch.php
+++ b/sources/ElkArte/AdminController/ManageSearch.php
@@ -321,7 +321,7 @@ class ManageSearch extends AbstractController
 			checkSession('get');
 			validateToken('admin-msm', 'get');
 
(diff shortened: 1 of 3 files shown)


Spuds:
Could you rebase this PR to the current dev ... as is its showing some old commits from 2018.  Thanks!

tinoest:
> Could you rebase this PR to the current dev ... as is its showing some old commits from 2018. Thanks!

What’s the easiest way to do that? I’m not good at rebasing. I tend to leave the commit history in place (as you can see)

emanuele45:
Well, I think that on a single commit it's easier if you pull "clean" the development branch => branch to a _temporary branch_ (optional, but easier on the long run because after you can just delete the temp-branch and pull again dev) => apply changes and push & send the pull request.

PR #3468, opened 2021-02-21
Changes to support fulltext on subject

tinoest:
Seems someone before me didn't use spaces, I'll sort those. 

Edit: seems I can't actually see them in my editor and changing them has no effect.


Spuds:
Thanks for making the PR to the dev branch as well

PR #3471, opened 2021-02-23
Check to ensure the result is an instance of mysqli_result

tinoest:
Fixes; https://www.elkarte.net/community/index.php?topic=5822.msg42134#msg42134

This should be clean as I force pushed over my old development copy


Spuds:
Thanks looks good !

PR #3472, opened 2021-02-24
Expose the (add/remove)Directory functions via the theme()->getLayers…

tinoest:
…()->getDirectory() call

This allow's modifications to still add an external Theme directory outside of the Elkarte theme folder structure, similar to how you can on Elkarte 1.1

### DIFF
--- a/sources/ElkArte/Themes/Templates.php
+++ b/sources/ElkArte/Themes/Templates.php
@@ -502,4 +502,12 @@ class Templates
 			}
 		}
 	}
+
+    /**
+     * @return Template Directories
+     */
+    public function getDirectory()
+    {
+        return $this->dirs;
+    }
 }


Spuds:
Appears like a good item to expose ... merged

emanuele45:
Why do you need to get the list of directories?
And anyway you should not get it from the templates but from the theme loader I guess...

tinoest:
> Why do you need to get the list of directories?
> And anyway you should not get it from the templates but from the theme loader I guess...

I wanted the current instance, that seemed the best way. I use it like this in TinyPortal https://github.com/Tinyportal/ElkTP/blob/50e67d905587ed69d80bc6b3ef801b59fff2b6ed/TinyPortal/Integrate.php#L563

PR #3482, opened 2021-04-04
Latest collection of 1.1.7 updates

Spuds:
- The current static array does not allow addons to supply their own unsubscribe method, this fixes that.
- ILA links in PBE were being double wrapped
- PBE code was being inserted in db and messages where it should not be (e.g. registration).  Harmless until you try to do an 1.0 to 1.1 upgrade since the db codes would be no conforming and the table update failed
- additional adjustments to pbe line length and base64 encoding codes
- the depreciated notice will really not be fixed here, ```octdec``` is likely getting a dec from the filesys and that makes it toss the error (or any whitespace as well which this does address)

### DIFF
--- a/sources/controllers/Notify.controller.php
+++ b/sources/controllers/Notify.controller.php
@@ -450,6 +450,14 @@ class Notify_Controller extends Action_Controller
 	{
 		global $user_info, $board, $topic;
 
+		$baseAreas = array('topic', 'board', 'buddy', 'likemsg', 'mentionmem', 'quotedmem', 'rlikemsg');
+
+		// Not a base method, so an addon will need to process this
+		if (!in_array($area, $baseAreas))
+		{
+			return $this->_unsubscribeModuleToggle($member, $area, $extra);
+		}
+
 		// Look away while we stuff the old ballet box, power to the people!
 		$user_info['id'] = (int) $member['id_member'];
 		$this->_req->query->sa = 'off';
@@ -472,6 +480,32 @@ class Notify_Controller extends Action_Controller
 				$this->_setUserNotificationArea($member['id_member'], $area, 1);
 				break;
 		}
+
+		return true;
+	}
+
+	/**
+	 * Pass unsubscribe information to the appropriate mention class/method
+	 *
+	 * @param mixed[] $member Member info from getBasicMemberData
+	 * @param string $area area they want to be removed from
+	 * @param string $extra parameters needed for some
+	 *
+	 * @return bool if the $unsubscribe method was called
+	 */
+	private function _unsubscribeModuleToggle($member, $area, $extra)
+	{
+		Elk_Autoloader::instance()->register(SUBSDIR . '/MentionType', '\\ElkArte\\sources\\subs\\MentionType');
+		$class_name = '\\ElkArte\\sources\\subs\\MentionType\\' . ucwords($area) . '_Mention';
+
+		if (method_exists($class_name, 'unsubscribe'))
+		{
+			$unsubscribe = new $class_name;
+
+			return $unsubscribe->unsubscribe($member, $area, $extra);
+		}
+
+		return false;
 	}
 
 	/**
@@ -491,7 +525,10 @@ class Notify_Controller extends Action_Controller
 	 */
 	private function _validateUnsubscribeToken(&$member, &$area, &$extra)
 	{
-		$potentialAreas = array('topic', 'board', 'buddy', 'likemsg', 'mentionmem', 'quotedmem', 'rlikemsg');
+		// Load all notification types in the system e.g.buddy, likemsg, etc
+		require_once(SUBSDIR . '/ManageFeatures.subs.php');
+		list($potentialAreas,) = getNotificationTypes();
+		$potentialAreas = array_merge($potentialAreas, ['topic', 'board']);
 
 		// Token was passed and matches our expected pattern
 		$token = $this->_req->getQuery('token', 'trim', '');
--- a/sources/subs/Emailpost.subs.php
+++ b/sources/subs/Emailpost.subs.php
@@ -133,6 +133,9 @@ function pbe_fix_email_body($body, $real_name = '', $charset = 'UTF-8')
 	// Remove the \r's now so its done
 	$body = trim(str_replace("\r", '', $body));
 
+	// Remove any control characters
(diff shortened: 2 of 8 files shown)


Spuds:
@emanuele45 Do you have any ideas on how to fix the ```octdec``` depreciation notice issue in ```elk_chmod``` I think what is occurring, and this is simply a guess, is in the areas where we don't send that function with say ```0777```  but rely on the OS perms value, it may be getting say ```33279``` decimal which is ```100777``` octal or ```0777``` as far as we are concerned.

emanuele45:
hmm...
the note on php page for octdec mentions 7.4, I tested few stuff on my machine with 7.4 but couldn't manage to get an error. I tried chmod with either the direct outcome of fileperms or even false, but no errors.
Also, I can't find the original report, do you have it around just to refresh the issue?

Spuds:
The only report was here: https://www.elkarte.net/community/index.php?topic=5914.msg42352#msg42352

In the PR I simply did a trim since any whitespace character is supposed to trigger the error, and I assumed pure integer as well but again w/o any real data it was just a guess.

emanuele45:
hmm... https://github.com/dompdf/dompdf/issues/2003 ?
Drop the double quotes around `$mode` in `if ($mode == decoct(octdec("$mode")))` ?

Spuds:
I tried it both ways, quoted and un-quoted, and it creates the error either way.   

I can trigger the error with any non octal value, so white space, alpha, or any  digit > 7.  That is why ```33279``` decimal which is octal ```100777``` causes the issue in my tests since it has a digit > 7

I think a fix could be
```
	if (($mode % 10) >= 8)
		$mode = decoct($mode);
	if ($mode == decoct(octdec("$mode")))
		$result = @chmod($file, intval($mode, 8));
``` 
That should catch any integers > 7, and then simply convert the number to octal.  The trim I added should be moved above the ``` if (empty...``` line as well.

ahrasis:
I did mention this also earlier https://www.elkarte.net/community/index.php?topic=5914.msg41898#msg41898

The error is still there as I tried your latest Patch_117_4-15-2021

Spuds:
Its a pretty frustrating issue really.

I've protected against empty strings and non octal values ... I  could add a ```is_numeric``` check as well since ```false``` or ```error``` would trigger the notice.  

But what is problematic is the error is being thrown pre-patch !  Example one of the errors you noted, the line number, post patch, is a comment line so it can't throw an error, but pre patch it is the suspect ```octdec()``` line.   So this means in some cases its throwing the error while its being patched, and I have no idea how to hide/prevent that at this time.

ahrasis:
But the fix doesn't work. Right after install I got this 
```
A non-numeric value encountered
```
for action=admin;area=packages;sa=browse; or action=admin;area=packages;

The packages page cannot be accessed totally.

Spuds:
Well that is a new behavior that I'll have to look into, I was trying to fix the error you had repotted earlier in this last update and looks like I made it worse Argh!  I removed the package since its clearly broken and needs to be fixed,

emanuele45:
@ahrasis maybe you can put some logging just before the call and see exactly what is the value that is sent to the function? Dunno if it makes sense, but at least it may help, something like:
```
	// Make sure we have a form of 0777 or '777' or '0777' so its safe for intval '8'
	if ($mode == decoct(octdec("$mode")))
	{
		Errors::instance()->log_error(var_export($mode));
		$result = @chmod($file, intval($mode, 8));
	}
```
What do you think @Spuds ?

Spuds:
Would not hurt ... maybe something simple like
```Errors::instance()->log_error('Mode :: ', $mode]);``` before that block of code?

**Comment by Spuds** [comment]
Just in case I was not clear, which is likely, 

The error(s)  about  ```Invalid characters passed for attempted conversion``` are from package.subs.php and UnTgz.class.php.   These are part of the package manager, as such the very act of installing a package can trigger the error.  

In our case the patch package, which *might* fix the error going forward, triggers the depreciation notice during the install since its still running the "old" code while its making the patches.  So a catch 22.  If we can capture the ```$mode``` value we can at least know if we have fixed the error, BUT it will still trigger one last time during the patch install.

**Comment by ahrasis** [comment]
The error is in the UnTgz.class.php line 392 after installing the said 1.1.7 patch.
```
		foreach ($this->_current as $key => $value)
		{
			if (in_array($key, $octdec))
			{
				$value = trim($value);
				if (($value % 10) >= 8)
					$value = decoct($value);
				$this->_current[$key] = octdec($value);
			}
			else
				$this->_current[$key] = trim($value);
		}
```
My guess is we need to check if the value is numeric as well in there.

For now, Package.subs.php doesn't log any error after adding the suggested lines.

**Comment by ahrasis** [comment]
These work in my test:
```
		foreach ($this->_current as $key => $value)
		{
			if (in_array($key, $octdec))
			{
				$value = trim($value);
				if (is_numeric($value)) {
    					if (($value % 10) >= 8)
	    					$value = decoct($value);
		    			$this->_current[$key] = octdec($value);
				}
			}
			else
				$this->_current[$key] = trim($value);
		}
```

ahrasis:
Any new / final 1.1.7 Patch release?

Spuds:
Not yet ... I need to add in your fix above and re-package it.  Other than that fix I really don't intend on doing any more changes.

ahrasis:
Seems good so with your Patch_117_4-15-2021 plus the above fix.

If all good with others as well, I think you all can already freeze 1.1 and then go for 2.0.

Sorry for not helping much. I have real life situations to handle especially for my NGO.

Spuds:
I was able to catch the error, at least one of them.  

It occurred when the updated ```cssmin``` library was being installed to the ```ext``` dir.  The exact error was that I was seeing is a decimal value of ```493``` which is not a valid octal due to the 9 (493 = 755 octal)

So the original "fix" of ```if (($value % 10) >= 8)``` would fix that *if* it was in place already ... but it is not changed until after the package is installed  (I think the error is actually thrown during the package test phase, which makes sense)

As a test I changed from using ```require-file``` where I was replacing the old with the new to a ```require-dir``` and placed the changed files in a directory.  That seems to prevent the error from occurring during the test/install of the patch.  Of course the patch has the above fixes as well since they are proper going forward.

The new patch is on the forum for testing.

ahrasis:
Ok. Will test that when I have time.

Spuds:
Thansk!
Your updates are also in the patch, as are the change to require-dir from require_file to help "hide" the issue during the upgrade.

PR #3483, opened 2021-04-05
Create Fulltext Subject Search

tinoest:
Modify the Subject/Topic only search to use a Fulltext index if using that Search Engine. 

This is far from tidy, I don't expect it to be merged and would like some feedback @Spuds @emanuele45 on what the better/best way to implement this is. It functionally works, but I don't personally like it.

I need to sort the limit functionality btw, I couldn't get my head around that at the start as it seems to not pay attention to the sort criteria and will ignore newer posts if the limit is to low.

### DIFF
--- a/sources/subs/Search/API/Fulltext.php
+++ b/sources/subs/Search/API/Fulltext.php
@@ -3,15 +3,15 @@
 /**
  * This search class is used when a fulltext index is used (mysql only)
  *
- * @name      ElkArte Forum
+ * @name	  ElkArte Forum
  * @copyright ElkArte Forum contributors
  * @license   BSD http://opensource.org/licenses/BSD-3-Clause
  *
  * This file contains code covered by:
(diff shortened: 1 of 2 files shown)


Spuds:
Is there a particular aspect that you are not happy with?  I need to pull this branch to really look at it since the github diff is not that helpful, at least to me!

@emanuele45 is the most knowledgeable of that area so it would be good to get his input.  I don't feel this will be merged for the 1.1.7 branch, mainly since I just want to be done with it.  Its fine for review, but would be a 1.1.8 item or certainly a 2.0 item.  I'm about burnt out on 1.1.7 😇

tinoest:
It’s messy primarily, I couldn’t work out the best way to pass in if it was a message or subject and message. 

It works functionally, just I’m sure could be improved.

Spuds:
Closing this request for 1.1.7 which is done and the branch needs to be removed.

I've cherry picked these changes for 1.1.8 so no need to re-open for 1.1.8

PR #3493, opened 2021-04-26
Munge117 part2

Spuds:
This should be all the changes, not applied separately, since the last 1.1.7 munge.  I'm sure I've missed somethings but yeah these two branches are far apart in certain areas so take what ya can get


Spuds:
Pass = merge !

PR #3501, opened 2021-06-30
Remove OpenID / Fix Tests

Spuds:
This removes OpenID support since its no longer supported at this version.  Surprising how many places this had its fingers in.

Also began to fix the web tests.  Appears some issue with mouse clicks not always working has shown up.  In working around that issue I tweaked a couple of templates to add selector ID's and cleaned up somewhat the install template.


emanuele45:
So long, and thanks for all the fish?
Can I push merge or you have other stuff to drop? :D

Spuds:
That is all for this PR ... be more in another :D

emanuele45:
:champagne:

PR #3502, opened 2021-07-04
Remove browser detect

Spuds:
This removes the browser detection class.  Most, if not all, of the needs were for ie < 9 workarounds, and some old opera ones, so those are no longer needed. The function was getting pretty crusty as well, lots of very old detection and missing new ones and some that are now invalid, basically I'm saying it was not really working as intended.

The one main check it use was the ```possibly_robot```,  That value was based off the user info value, so I simply used that in the two places it was called out.  We may need to add back the "not a robot" if you are logging in or registering, but I could not find how that change to the value (via the browser class) was actually used.  If it is need, it can be added back in without the weight of that class.

There is one sniff test left in the php code, and that is to turn off the search dropdown, we can't simply hide that on small screens since the code alters the scope of the search and w/o the dropdown you would not know.  That code should be updated IMO, I think it too often confuses people when using search (that the scope was automatically limited to a topic or board)

Since I had to look at the robot detection stuff, it was again pretty old, so I added in a few and removed a few, based on the top ten bots per region.

The browser detection in the JS was also updated, but the few area it was used were no longer needed.  Really its only the mobile / touch stuff.

Fixed a few things found while testing all of this.


Spuds:
Yeah ... I pissed off maxPain !   To look into

emanuele45:
> Yeah ... I pissed off maxPain ! To look into
So mean...

Spuds:
Ha take that Mr. Mean!

Looks like there may have been a couple of hidden fails, phpunit is strange like that, those have been fixed.  The others were, shockingly, caused by me.  

I changed the ```httpreq constructor``` to be private to force the use of ```::instance```
 -  (something we should do one any other class constructors that we expect to use ::instance)

This caused a couple of other tests to fail that were directly calling ```new httpreq()``` and then trying to stuff $_GET ... neither is cool really

Anyway all fixed.

Spuds:
I've heard no objections .... in it goes!

PR #3504, opened 2021-07-11
Login / Registration forms

Spuds:
A few things here.  

-  removes the user entered session time, simply allows "forever" or uses what the admin has in the ACP.  No changes to the cookies themselves in this update
- uses some experimental form inputs for text/email/password drawing from what material design does.  The concept is to keep the input labels close to the box.  Login was not bad but registration was a mess.  You can test drive at elk.spudsdesign.com by looking at the login and registration tabs.
- As always various fixes to areas that showed up in trying to test the changes or misses from my previous commits 👼


Spuds:
I've left the crowd speechless in the wake of this awesomeness ... merge!

PR #3525, opened 2021-12-26
The followup parameter can come from both a GET or a POST, so... getRequest - fixes #3497

### LINKED ISSUE #3497: Followups Not Working
**Describe the bug**
Use "more" dropdown and select "quote to new topic".

You are redirected to a new topic post page as expected but w/o the quoted message.  Go through the posting process and its not a followup as it should be.

**ElkArte info**
1. 2.0

emanuele45:
That was kind of easy. :D

### DIFF
--- a/sources/ElkArte/Controller/Post.php
+++ b/sources/ElkArte/Controller/Post.php
@@ -427,7 +427,7 @@ class Post extends AbstractController
 		$ns = $this->_req->getPost('ns', 'isset', false);
 		$notify = $this->_req->getPost('notify', 'isset', false);
 		$quote = $this->_req->getRequest('quote', 'intval', 0);
-		$followup = $this->_req->getPost('followup', 'intval', 0);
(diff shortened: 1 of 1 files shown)


Spuds:
Looks like one of many issues caused by my $req implementation.  Thanks for the find!

emanuele45:
@Spuds should I care about the maxpain tests failing?

Spuds:
I had pushed a fix (yesterday) that fixed the tests (or at least should have).  There were a few issues, one was pointing to the elksite which should have been easy to remedy, but it reveled a couple of other issues with our stream and fopen handlers that I fixed.

emanuele45:
Rebased and pushed.

Spuds:
Yeah, I fixed this again in another PR.  Its working but the search results, for some reason, changed a bit.  I think it was some odd regional result that I've tried to stabilize in the fix (set language etc).  Anyway going to merge this as the test error here will get sorted on its own.

PR #3531, opened 2022-01-03
Updated Db-PostgreSQL to be PHP 8.1 compatible

tinoest:
Changes to make 1.1.8 compatible with PHP 8.1 

@Spuds @emanuele45 there will be other changes needed strftime/gm_strftime is another error, do you want those in a different PR or is this one ok?

### DIFF
--- a/sources/database/Db-postgresql.class.php
+++ b/sources/database/Db-postgresql.class.php
@@ -1074,6 +1074,11 @@ class Database_PostgreSQL extends Database_Abstract
 		return (is_resource($connection) || $connection instanceof \PgSql\Connection);
 	}
 
+	/**
+	 * Return an id for the resource/object
+	 *
+	 * @param resource|object $resource of returned result set
+	 */
 	public function seekCounter($resource)
 	{
 		global $db_row_count;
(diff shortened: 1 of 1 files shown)


sources/database/Db-postgresql.class.php
```diff
@@ -1064,6 +1071,20 @@ public static function db()
...
 	public function validConnection($connection = null)
 	{
-		return is_resource($connection);
+		return (is_resource($connection) || $connection instanceof \PgSql\Connection);
+	}
+
+	public function seekCounter($resource)
+	{
+		global $db_row_count;
+		static $prev_id = null;
```
emanuele45, line 1080: Nitpicking: not used?

emanuele45:
If you want to keep this one specific to psql then you can create other PRs and the other changes are not related to psql then sure create new ones.
Otherwise if you want to change that to "php 8.1 compat", then fine here too.

tinoest:
> If you want to keep this one specific to psql then you can create other PRs and the other changes are not related to psql then sure create new ones. Otherwise if you want to change that to "php 8.1 compat", then fine here too.

Resolved those issues, nitpick away. Although I hate the { } on newlines. 

```
if (condition) { 

}
else {

}
```

Is what I do so I'll just have to remember you put them on newlines. I'll create a new one for the date stuff as it's going to be a fair few changes.

tinoest:
Changed base to 1.1.9 instead of master as per #3538

PR #3532, opened 2022-01-03
Site triage and 1.1.8 pull in

Spuds:
This also brings in the 1.1.8 changes, including the security issue, should have been a separate PR but that is not the way this worked out.  Better next time.  To note, it does not include the search changes from 1.1.8 as the 2.0 search code is so different that it was not possible to resolve conflicts, it needs to be done manually.


Spuds:
And I'll fix the tests again LOL

PR #3533, opened 2022-01-03
PHP 8.1 strftime and gmstrftime changes

tinoest:
@Spuds @emanuele45 

Not so sure this the correct way to go about this or if it should be in Util:: I think Util makes more sense? 

It makes no changes to the code if you are before version 8.1 (we might aswell use the old functionality) newer it tries to do a replace on the text string to a format date can handle. I might expand it out and use date_create date_fromat etc. As that seems to handle locale's better.


Spuds:
Thank you for moving those functions to Util:: ... agree it makes a more sense, and allows for easy addons use.  I don't currently see any issues with the PR,  let me know if you are done 😄  

Also I'm assuming this is required in 2.0 as well?  (I'm in the middle of some other 2.0 stuff so have not yet looked)

tinoest:
> 

There are issues with Elkarte 2.0 on PHP 8.1 above this, I am currently working through them and can submit a PR with those in also when I am done.

I need to test this with a timezone difference as I think there will be bugs, but once I have done that it will be ready to be merged, you can always do that now and I'll follow up with another PR with those changes in if any are required. This change should be transparent to anyone less than PHP 8.1 anyway, other than an additional function call.

tinoest:
You can merge this I think. Associated PR for Elkarte 2.0 is #3537

Spuds:
Awesome ... Thank you 👍

Spuds:
@tinoest Note I reverted this since it was a direct to master, please make the PR to the 1.1.9 branch.

PR #3537, opened 2022-01-07
Postgres and PHP 8.1 Fixes 

tinoest:
Updates for Postgres and other associated PHP 8.1 changes


Spuds:
Looks like the test failures are due to the 2.0 namespacing ... some of those files will need a ```use ElkArte\Util;``` added to the top

tinoest:
Ok, I’ll sort them tomorrow.

tinoest:
I’m not quite sure what it’s failing on now?

Think I’ve caught all the missing namespace issues.

Spuds:
Its a false fail, you are all clear.  I'm still working on stabilizing a couple of fetch web data test (with post data)

PR #3539, opened 2022-01-07
Strftime fix

tinoest:
#3533 - Against 1.1.9


Frenzie:
Thanks! I just glanced at the error log after trying PHP 8.1 and wowzers. ;-D

PR #3540, opened 2022-01-09
Moving language files out of the themes directory

emanuele45:
First pass to see if tests still work.
I don't entirely like it (as expected) but it's _something_. :smile:


Spuds:
HA ... those test fixes are on you 🏃‍♂️

Spuds:
Is this still under 🚧  ?

emanuele45:
Yep, yep!

emanuele45:
I'm currently in the phase: should I go all out or should I refrain a bit and try to avoid some overhead? :angel:

Spuds:
I don't think you are really asking 🤣 Its 2.0, if you don't do it now, well its unlikely to ever get done.  Just sayin'

Spuds:
Conflicts .... Your welcome very much 😸

emanuele45:
That was easy. :stuck_out_tongue:

emanuele45:
@Spuds opinion: I'm currently unsure again about the scope. What I plan to do right now is to remove all the part that allows to edit files to change strings and put in place a db table with the changes (so that we distribute the php files and if an admin wants to change a string, that is stored in the db and will override the one in the php file at run-time). Kind of easy-peasy.
Then, while moving stuff around and doing the changes I made to the string loading code, I started wondering if we could do something that doesn't have to directly rely on php files, and one option I was wondering was to distribute JSON files with pair `key:text` and convert them into cache entries for faster loading. I was also thinking about gettext, but maybe it's too much.

Though, I'm not sure if it is actually worth the hassles. :thinking:

Spuds:
I really like that easy-peasy idea !!

I'm not sure the conversion of the PHP to JSON would be worth the effort, thats not saying it would not be better, since it would,  not clear how much loading would improve.

I was going through some of our strings and we seem to have quite a few duplicate values under different names and some what I would consider standard values that should be in index but are located elsewhere.  I think we could do some good consolidating them ...  Maybe if index.language.php was more robust / complete we could consider a json and cache thingy just for it?  Again not sure, just typing out random thoughts.

emanuele45:
String consolidation is always a bit tricky because of the different languages that sometimes do not behave like English, but that's string-by-string.
But indeed some duplication is very likely to be worth removing.

The JSON VS PHP was not really for speeding up the loading in comparison to now, actually I'm pretty sure it would be slower.
The idea I had was more related to how the variables are set.
In current-elk we just load in global and that's it (leading to some rather nasty swapping and reloading and stuff that is kind of hackish).
In this branch I took the opportunity to try to avoid the global and relying on the variable scoping of PHP isolate the loading of the variable into the function that does the require of the language file, for then assign its value to the global. Still not exactly clean.
In the code I was thinking, the cache file would contain a function that returns the array, that way the code that loads the language file would "just" merge it into either the global or a property of the class or whatever.
That *should* be a tad cleaner and avoid things like: 1) the "fix array" for the days of the week, 2) the check for $txtBirthdayEmails that is not exactly nice, and 3) the naming of the variables for the email templates that could then go back to a proper array like it originally was (and that has been "flattened" in order to adapt to transifex that doesn't cope well with nested arrays).

Problem is, despite these advantages, I'm still not sure it is worth the hassle. :unamused:

emanuele45:
The main stuff is done, but before merge I have to deal with agreement and privacy policy and the variables at the beginning of Index that are still not properly handled during the save, even though I'm mostly thinking of just not allow the editing.
I mean:
```
$txt['lang_locale'] = 'en_US';
$txt['lang_dictionary'] = 'en';
$txt['lang_spelling'] = 'american';

// Ensure you remember to use uppercase for character set strings.
$txt['lang_character_set'] = 'UTF-8';
// Character set and right to left?
$txt['lang_rtl'] = false;
// Capitalize day and month names?
$txt['lang_capitalize_dates'] = true;
```
What do you think @Spuds ?

Spuds:
is ```$txt['lang_spelling']``` used anywhere but pspell?  If so I'd say its probably time to remove those functions, does anyone use that anymore?  Its a pre browser spell check legacy.

Some of the others can really Fup things if set wrong, so not allowing the edit is probably the best option.

emanuele45:
Spuds wrote on 1/23/22 12:12 AM:
>
> is |$txt['lang_spelling']| used anywhere but pspell?
>

Is used in pspell both for spellchecking posts (that is indeed... not 
really a feature worth of keeping nowadays I guess) and in pspell_new in 
search for suggesting correct spelling (I suppose).


> If so I'd say its probably time to remove those functions, does anyone 
> use that anymore? Its a pre browser spell check legacy.
>

Yup, the part about spellchecking posts is rather uninteresting to me.


> Some of the others can really Fup things if set wrong, so not allowing 
> the edit is probably the best option.
>

Cool.
Gone.

emanuele45:
I **think** that's all.
Finding all the places where that stuff is used is tricky. :-\

If you see/remember/know any place I may have missed, of course let me know!
TBH I didn't test much the last few commits (those related to the themes and agreement) mostly because of my weird setup that makes testing a bit annoying.

The only remaining issue in the current set up is that during future upgrades we may end up overwriting the privacy policy and the agreement. The option could be something similar to Settings.sample.php, and renaming them while we do the upgrade... that could work, yes.

Spuds:
:cool:  Hope you don't mind if I do not troll through 221 changed files 👯 I pulled this in and did some light testing and it seems to be working fine.  I'm going to pull this in and then rebase and conflicts  on a couple of PR's I'm going to make.  After this and what I have coming, probably time to refresh the site.

PR #3555, opened 2022-01-14
Reference User correctly

tinoest:


Spuds:
Instead, just place a ```use ElkArte\User;``` at the top of the file with the other ```use``` items ... its the new require once 😸

tinoest:
I did think that, then looked in other files and you’d referenced it that way. So went with Elkarte\User 😅

Is your preference an use?

Spuds:
LOL ... yes the preference is the ```use``` statement all around.  Some large changes  were done with *cough* copy/replace *cough* and that is likely why there are some inconsistencies. This time I'm 😇

PR #3562, opened 2022-01-19
Next batch of fixes

Spuds:
As seen or reported on the site


Spuds:
If there are mistakes in here, rest assured there are still more to go in the code, so lots of opportunities to fix

PR #3568, opened 2022-01-23
Make sure only return is an accepted type

tinoest:


Spuds:
Thank you, was just about to make those changes and noticed this PR

PR #3573, opened 2022-01-26
Couple of language loader typo's

Spuds:
@emanuele45  I noticed a couple of typos, here is a quick PR.


emanuele45:
So unlike of me, I'm usually not into typos. :stuck_out_tongue:

PR #3574, opened 2022-01-27
Next Batch of Fixes

Spuds:
Another collection of fixes, tweaks and enhancements

- added support for tiktok vids to the embedder.  I don't watch but they are popular enough to support.
- fix broken email links
- fixed several more geturl links that got a bit munged
- added new SVG loader/spinners that do not require CSS keyframes
- redid the progress bar so it has less markup
- fixed several ajax buttons that were not working
- fixed as indicated reported issues.
- any spelling or casting or formatting that got in my way during the above


Spuds:
Time to Let Er Rip Tater Chip

PR #3575, opened 2022-01-28
Mark as read controller

Spuds:
This is an attempt to redo the mark as read controller which was previous hot wired to work with API calls and other mystery.   

I'm not ecstatic with where it is, but I feel its now workable.  It properly passes though the action controller as a "normal" controller, and handles the API calls through the normal flow and not injected to a random method.  Still lots of redirect exits and return deliciousness to keep everyone guessing.  Better yes, great no, at least its no longer in the action array!

Also fixed the JS since I had not updated that when I changed all the CSS (Note these changes rely on the fix3 commit changes as well, respectably the CSS changes to boardindex, I should have had them in this PR but sometimes that is now how things roll 🎲

### DIFF
--- a/sources/ElkArte/Controller/Markasread.php
+++ b/sources/ElkArte/Controller/Markasread.php
@@ -127,7 +127,7 @@ class Markasread extends AbstractController
 		}
 
 		// Dispatch to the right method
-		//$action->dispatch($subAction);
+		$action->dispatch($subAction);
 
 		// For the time being this is a special case, but in BoardIndex no, we don't want it
 		if ($this->_req->getQuery('sa') === 'all' || $this->_req->getQuery('sa') === 'board' && !isset($this->_req->query->bi))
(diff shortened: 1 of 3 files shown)


emanuele45:
TBH I forgot why I did it that way. :joy: 
Probably I was having the idea that since the controller was accessed basically only by AJAX-like calls, force the API was reasonable. But indeed it forced some ugly stuff. :unamused:

Spuds:
Just noticed I left a debug comment in, so I need to send in a quick fix for this or it will do nothing LOL

Spuds:
The way the API stuff worked was (kind of still is) a bit wonky.  It used to have the short cut in the dispatcher to call _```api ```methods directly.  I removed that as it was a) hard to trace/debug and b) really opens potential security issues.  We also had things like ```;api;xml or ;xml or ;json``` and other combos.  Now I think it only supports ```;api=xml or ;api=json``` type things with a method to see if something is/was an api request and what is expected.  I just felt we had to much juggling (and still do) in the controllers for vague api requests.

PR #3580, opened 2022-01-31
[WIP] Attachments

emanuele45:
It was a while I wanted to kill again the attachments.
That's really just a bit of makeup to warm me up. The Attachment class is a placeholder, I just wanted to put these constants somewhere because it always bothered me I could never find what the heck the numbers were. It could hold much more, worst case a good chunk of Attachments.subs.php in a static way.

I was thinking of a couple of things:
1. drop the storing of the thumbnails in the db (just rely on a prefix to the file name and GTFO),
2. *maybe* move the uploaded avatars out of the way (that as far as I remember is a good chunk of annoying code to deal with),
3. if not too messy, maybe go back to when there was one way to store attachments, just not a single directory, I'd say by months is usually good enough and easy to deal with?

What do you think @Spuds ?

### DIFF
--- a/sources/ElkArte/Controller/Attachment.php
+++ b/sources/ElkArte/Controller/Attachment.php
@@ -286,8 +286,7 @@ class Attachment extends AbstractController
 		// We need to do some work on attachments and avatars.
 		require_once(SUBSDIR . '/Attachments.subs.php');
 
-		$id_attach = $this->_req->query->attach ?? '';
-		$attachment_class = new Download($id_attach, $this->_req->getQuery('attach', 'intval', 0));
+		$attachment_class = new Download($this->_req->query->attach ?? '');
 
 		$inline = $attachment_class->isTemporary();
 
--- a/sources/ElkArte/Controller/Attachment.php
+++ b/sources/ElkArte/Controller/Attachment.php
@@ -287,7 +287,7 @@ class Attachment extends AbstractController
 		require_once(SUBSDIR . '/Attachments.subs.php');
 
 		$id_attach = $this->_req->query->attach ?? '';
-		$attachment_class = new Download($id_attach, $this->_req->getQuery('attach', 'intval', $this->_req->getQuery('id', 'intval', 0)));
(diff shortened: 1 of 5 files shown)


Spuds:
WooHoo More attachment fun 👍 

1) Not storing the thumbs in the db ... I feel this makes sense.  I think the code still does a fileexists after the db call so why not just look for the attachment hash + _thumb.  Should save some overhead I think.

2) I abused the avatar code in a recent commit (no more uploads as attachments, removing all those db calls).  Moved all of the avatar upload code out of profile into its own class, and more so just unraveling that messy code.  It really did not do what it claimed to do.  Anyway any improvements there are welcome.

3) You mean drop the auto directory options and just make it year/month going forward.  That should simplify several areas.  I feel some of the current options are a bit confusing and make the UI difficult to follow.

emanuele45:
Dealing with attachments is more of an headache than I remembered. :laughing:

1. yep, apparently they are already saved with "_thumb_" but stil lalso in the db with the whole id thingy... confusing.
2. I saw, didn't notice at first!
3. yep, I wanted to do it a while ago but gave up. The only downside is that we'd have to reprocess all the attachments during the upgrade... that may become a bit time consuming...

tenbuddy:
It's an interesting matter that I learned many things from here. Everyone is so co-operative and I would love get the [latest news]([url]https://www.oracleglobe.com/(url)) update.

sources/ElkArte/Controller/Attachment.php
```diff
@@ -286,8 +286,7 @@ public function action_dlattach()
 		// We need to do some work on attachments and avatars.
 		require_once(SUBSDIR . '/Attachments.subs.php');
 
-		$id_attach = $this->_req->query->attach ?? '';
-		$attachment_class = new Download($id_attach, $this->_req->getQuery('attach', 'intval', 0));
+		$attachment_class = new Download($this->_req->query->attach ?? '');
```
live627, line 289:
```suggestion
		$attachment_class = new Download($this->_req->query->attach ?? 0);
```

Spuds:
Some/Most? of this has been added in one form or another in other PR's

PR #3585, opened 2022-02-05
Emoji Tweaks

Spuds:
Perhaps just for my personal sanity and to take a load off git whcih was moving at a crawl on my old computer.

This changes the emoji SVG's from being fully unpacked in their (3) directories to now only being unpacked when chosen.  If you change which set is in use, it will clear the previous set so only one pack is expanded at any one time.  So instead of 3.3K+ files there are now 3.

I also (properly) fixed 🙏 the Fx issues where it was calling the emoji class before the data file was loaded.  Chrome was fine but Fx did not wait for the promise to complete before calling the next sceditor init process.  Anyway, small tweak and it seems all good now.

Yeah thats right, 3,382 files LOL


Spuds:
Merged this in so we would no longer have to deal with 3.3K files switching branches etc.  Will cause a slow pull or rebase but so worth it after that :D

emanuele45:
@Spuds I'm having a little trouble with this commit, after I pulled it I got the message:
```
warning: CRLF will be replaced by LF in smileys/noto-emoji/noto-emoji.zip.
The file will have its original line endings in your working directory
diff --git a/smileys/noto-emoji/noto-emoji.zip b/smileys/noto-emoji/noto-emoji.zip
index 19854c9aa..4fd536e70 100644
Binary files a/smileys/noto-emoji/noto-emoji.zip and b/smileys/noto-emoji/noto-emoji.zip differ
```

And even if I try to commit it I get:
```
fatal: CRLF would be replaced by LF in smileys/noto-emoji/noto-emoji.zip
```
Any idea? :thinking:

emanuele45:
And, if I try to open the file with Ark i get the error:
```
Loading the archive ‘/.../smileys/noto-emoji/noto-emoji.zip’ failed with the following error:
Failed to open archive: Zip archive inconsistent: central directory overlaps EOCD, or there is space between them
```
:raised_eyebrow:

Spuds:
HUmm ... They are all doing that, looks like git did not seem the .zip as a binary and F'ed them all up!   I'll have to push them again.

Spuds:
Updated the ```.gitattributes``` so it treats archives as binary.  Also refreshed the zip packages (should have done that when I made the PR) to update to the latest tweaks from the various emoji repos and pushed that.

PR #3594, opened 2022-02-13
Add WebP support

Spuds:
This adds webp support to attachments, avatars and resizing.

- Option to allow webp added to the ACP, system also checks for GD or IM support.  When enabled
- Oversized avatars will be saved as webp.  Normal sized avatars processed normally, including webp ones
- Supports normal webp attachments thumbnail generation
- Attachment thumbnails will be saved as webp
- Resized attachments *might* be changed to webp if reduced size is needed.

Cleaned up some formatting, and universally call ```canUse():``` for graphic support checks: 

One thing I changed was https://github.com/Spuds/Elkarte/blob/2094f7a6deebbe86a0afa93771aa7d71b973804a/sources/subs/Attachments.subs.php#L1187 
@emanuele45 it looks like that has been in there from the start and I don't understand why it was set that way, so maybe you do?
To me it It says create a thumbnail if 
Its not been created (ok) 
OR its been created but it is to wide (ok) 
OR its been created and its to high (ok) 
OR its to both under the width and height limits (???)  
What it was doing is if you have say a small avatar, one just under the WXH admin limits, it will try to recreate the thumbnail, but of course thumbnails are set to never scale up so it just rolls on through, perhaps many many times as well.  Seems like it was to fix or prevent something but ??


Spuds:
Accidently deleted the branch 🤦

Spuds:
I merged this in so it can be further tested.  I'm still not clear on
https://github.com/Spuds/Elkarte/blob/2094f7a6deebbe86a0afa93771aa7d71b973804a/sources/subs/Attachments.subs.php#L1187 why that logic was like that.  So @emanuele45 if you have any ideas let me know, its just commented out FTM

PR #3595, opened 2022-02-14
Prevent archive files from being corrupted

Spuds:


emanuele45:
That seems to work fine, great! Thanks!

PR #3597, opened 2022-02-14
Quick Reply Tweaks

Spuds:
This is inspired by the QRR addon by @ahrasis done in a 2.0 fashion.  It moves the quick reply area to the bottom of the message list and makes it look like a post area.  See picture.

image


emanuele45:
Pretty sweet!!

Spuds:
🆒 glad you like it!

PR #3602, opened 2022-02-17
Remove theme editor thingy

Spuds:
This removes the ability to horribly mess up your site by making CSS and Template edits with the Admin Control Panel.

@emanuele45 what do you think ? (specifically about this 😄 ) It was old inherited legacy code that really was very clunky to use, if it even did what you expected.


emanuele45:
It should have worked.
TBH I never used it, no idea how useful or not it may be. :man_shrugging: 
I guess even only editing the CSS nowadays is not an easy thing to do in a form...

Spuds:
I still miss the :fire: guy ... 

If you want to keep it, then "you own it" so to speak 😸 I think the function is klunky to use, open to multiple security issues, and an excellent way to crater you forum since you can edit the ACP stuff and do just that.

The other thing I *was* going to work on was removal of the FTP functions for the chmod controls for several reasons.  First the code is almost ensured to trigger a fail2banand lock the admin out of the forum for some period (it does on my server at least).  Its also not terribly secure and can leave your system a bit defunct due to some of the cluge in there.  Seems like a tack better left to dedicated FTP programs or a users web control panel.  I'll shelf that task since I know your an FTP lover,  but I'm not going to look at how to fix the issues in ElkArte since I simply am not interested in that function 😇

emanuele45:
lol

No, I was not saying to keep it!
My last sentence was "even editing the CSS nowadays if not an easy thing to do in a form on a web page, so the editor is pretty much useless". 😸

emanuele45:
I just don't merge because github doesn't like my browser. :D

PR #3603, opened 2022-02-18
Inconsequential tweaks and minor bug fixes

Spuds:
Cleaning up some branches ... tests will likely fail due to DDG 403 ... fixed in a PR elsewhere so 🤷 see ya!

### DIFF
--- a/.github/LICENSE
+++ b/.github/LICENSE
@@ -1,6 +1,6 @@
 BSD 3-Clause License
 
-Copyright (c) 2021, ElkArte Forum contributors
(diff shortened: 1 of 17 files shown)


Spuds:
Confirmed that the failures are from our old friend fetchweb ... seems DDG changed something (those test are fine from my local but not from the github vm) .... I fixed those tests in another PR but that has not been merged

Spuds:
Rebased onto the base which now has the test fix .... Now it should pass 😸

PR #3605, opened 2022-02-27
Packman

Spuds:
This is a pass through packman, intended to fix a few areas.

I pulled out the ```create_chmod_control``` to its own class as having that mini beast in package.subs was a bit crazy.   Going through that I found a rew rough spots and tried to clean them up while doing the refactor.  This was really 3 functions in one and although in the same sub, were essentially isolated with if/else stuff (meaning there was no shared code anyway)  Now that is in sources```/ElkArte/Packages/PackageChmod.php```

I also pulled out the ```parsePackageInfo``` function to its own class, again this was stuffed in package.subs.   This was one of those heinous massive if/elseif/elseif/elseif ...... constructs.  Moving it to a class allowed the removal of code duplication and other misery,  All of that is now in ```sources/ElkArte/Packages/PackageParser.php```

Lastly I did some updates to the ftp controller which was broken if you wanted a secure connection.  I also moved the chmod logic so it can choose f 775 or 777 is the right thing to do vs if/elseif/maybe this/or that constructs that were in a lot of places.  Now you call it with [755, 775, 777] and it will try each until success or fail (like what is in the file functions class now) 

OK so should you wan to test, try some FTP stuff.  I force tested that but I don't have a good, say apache suExec machine to verify on.  Also note the "reset" permissions function is somewhat broken in 1.1 since its on the redirect page, so the page redirects away from the form asking to reset, somewhat harmless, but could annoy some more security minded .

Several other areas affected in order to call the (2) new class functions.  Also MOVED all package functions to their own namespace ```\ElkArte\Packages``` as there are several files now.

Up next is to **REMOVE** the ftp permissions section in the ACP,   This is the one where you can apply, or choose your own, set of permissions to various files.  This function is simply dangerous and we should no longer have that in our product.  If a user needs to 'F' around like that, then they can do it on their own.

So much more can be done with pacman, but I want to move on to another "fun" area 👯   @emanuele45 let me know what you think and feel free to make any changes you want, and if you have a good server to do some FTP chmod testing that would be great.


emanuele45:
Only reading the comments make my head spin faaaaaaaaaaaaaaaaast! :laughing:

I'll try on a shared and see if makes any difference.

Spuds:
That would be awesome.  I started and stopped on this one several times as I kept getting lost LOL.  My setups are all fpm nginx so I had to force the need for ftp access (the fpm uid/gid is that of each user directory where the files arer)

Anyway, the main goals were to extract / unwind the ```change_mod_control,``` extract / unwind the ```parsePackageInfo``` bits (test/install) which were somewhat intertwined with that change mod.  Lastly to try and clear up the use of ```fileFunctions``` vs ```ftpControls```. so they acted similar to the code and push all the "try 755 no? try 775 no? try 777" logic to a single place vs multiple times throughout the code.

emanuele45:
hmm.. little weird.
I installed on a shared hosting, but had troubles because the server reports mysql 5.6.45 so elk didn't want to install (changed the limit and installed LOL).
Then, the files permission page doesn't seem to work, but based on your comments it may not be worth the hassle fixing it. I tried installing a sort of fake package (an easy one I could just hack somehow to maybe install), but at first try it gave me a "some path is not writable" but without details so I just went back hoping to get the same error again, but I didn't and from the second it gives me a "parser error" (see image, if it loads, otherwise https://i.imgur.com/RgRHGLC.png ).

Spuds:
The parse error could be file permissions?  Did it ever ask for FTP credentials?

TBH I don't know what the lower limit might be for the various DB engines'

Spuds:
I attempted to setup an apache / mod-php / suexec so I can try and help debug my mess !  ... I think I have something working correct, not sure, but now I remember why I don't use apache anymore.  

PHP is running under ```www-user:www-user``` and my forum is under ```web1:client1``` The apache user is also part of the ```client1``` group.  So 755/644 on the forum path work fine but PHP does not have write access and will need some FTP.  Sound right? 

---

 The first rather grotesque error (not mine).  
 - When the FTP connection is established ```getFTPControl()```, we determine the FTP root path (assuming here that the FTP user is chrooted, which on shared they should be)
 - In this setup my web root / BOARDDIR is ```/var/www/clients/client1/web1/web``` which is a path construct that a control panel like ispconfig would create, other panels create similar paths.
 - When I login with FTP I'm chrooted to ```/var/www/clients/client1/web1``` as my root and my path to the forum is ```/web``` or at least that is what I entered into the form,  simply ```web``` would work also (e.g. CWD web or CWD /web)
 - The system checks that we can login with FTP, then does a chdir command to check the path and sets an error if its wrong need to test).  It does some auto corrections in case you are using a ```/home/me/public_html``` setup from the 80's jk
 - Next it does a ```$ftp_root = strtr(BOARDDIR, array($path => ''))``` using path ```/web``` and the result is ```/var/www/clients/client11``` Brilliant! Not only the wrong path but now trying to access another clients path !  
 - Using just ```web``` would get you ```/var/www/clients/client1/1/``` equally wrong but at least in your path.

I think it may be best to just do a ```substr(BOARDDIR, 0, -strlen($path));``` If the CWD was successful for $path that would at least get you ```/var/www/clients/client1/web1/```  its important that root path be correct or all FTP commands will fail.

Spuds:
@emanuele45 https://github.com/elkarte/Elkarte/pull/3605/commits/2a941e7855d8973715b550a5c1911d591b2f2b7e is the commit that removes the File Permission change-o-rama from the admin panel, 

There was a lot of code in that, PHP/JS/Templates, I think I got it all but not sure.  I wanted to place it all in one commit in the event that you feel strongly that we should keep that function.  As it stood the JS was a bit broken and I feel the function itself was not as secure and well implemented as it should be.  It was/is just more than I wanted to undertake given the real usage of it in the wild.

I made a couple of changes to the other packman functions, seemed to work on my cobbled up install, but I was not able to duplicate the issue you showed.   Also note that the FTP functions are only 100% working with IPV4, some IPV6 (anything that wants to enter passive mode) will fail (mostly silently)

emanuele45:
Spuds wrote on 2/27/22 3:36 PM:
>
> The parse error could be file permissions? Did it ever ask for FTP credentials?

Sorry for the late reply.
oh crap. The error was just a file name error. :facepalm:

hmm.. I changed folders permissions to 555 but apparently the addon installed anyway. :thinking:
What would you suggest for testing at that point?

Spuds:
For FTP,  the webserver has to be running under a different user/group than the site files.  If they are the same then standard PHP calls will be able to fix the issue, like that 555 was likely changed to 755.  The FTP route does not reset them nor ask to reset them since that would be senseless.

I was able to do some tests by changing the site files to a different owner id than the web server but with same gid group as the then the web server.  Then setting up an FTP user with that id (jailed to the site file directory), then 755 should fail and require a 775 which that FTP user would be able to do.  

I also did some with testing with PHP in old cgi mode which runs under a different uid than the site files. 

Anyway the FTP testing is a pain in the rump LOL and I don't think most modern setups really require it (hence the issues with getting a setup to need it)  I don't know if I'd worry to much about it at this point, that old ftp class does not work with ipv6 nor does it properly connect with a sftp server, so its limited and insecure.  

I may fix the sftp stuff, I tried fsockopen and streams and could get neither worked, but built in ftp functions did work.   I think we should consider redoing that entire ftp class to just use built in ftp controls, and if your PHP install disabled them simply tell the user to do it on their own.

emanuele45:
So I guess we can take the easy path: call it a day and let others test it. If it works for most of the IT industry, it should work here too. :laughing:

Spuds:
Sounds like a plan ... really the next thing is to update the ftp client function which is ipv6 and TLS deficient.

PR #3610, opened 2022-03-06
Mail

Spuds:
This is an update to our `sendmail` functionality.  Sendmail was stuffed in `mail.subs.php` and had quite a few paths through the code along with some recursion etc.  

This PR breaks out (3) new classes to deal with what sendMail previously did.

- `BuildMail` is a class that deals with crafting the proper email headers, handling any special PBE needs.   It also builds the actual email message bodies.   The (3) bodies consist of:
  - Plain text.  This is an ASCII only version as required by mime multipart.  This has control characters removed and any >127 ASCII characters are converted to `#&123;` entities to remain complaint.  If HTML was supplied it is stripped out, although it adds some newlines at the block level elements `(</div> => "\n")`.
  - Encoded text.  This is a `base64` encoded version of the message.  Any entities are changed into proper utf-8 characters then the result is base64 encoded always.  The old function would only base64 if >127 ASCII characters were found but I feel this is a unnecessary distinction and simply adds another execution path.  This is set as text/plain
  - Encoded text.  This is a `quoted-printable` version of the message.  It undergoes the same processing as above but the result is then QP encoded.  If HTML is provided it will set it as text/html otherwise text/plain.  This section would not appear at all in the old function unless you set and supplied HTML -or- if it was doing a hotmail fix.
 - The `hotmail` processing used only for `at&t, bellsouth and comcast` has been removed.
---
- ```Mail``` class deals with the sending of what was created in `BuildMail`.  This will use php `mail()` or `SMTP` based on ACP settings.  Previously the `mail()` path was inside of sendmail and `SMTP` was its own function.  Lots of duplicate code between the two has now been removed and several issues resolved.  This function also takes care of the PBE code injection in to the various message bodies.
---
- ```QueueMail``` breaks out the `reduceQueue` function.   Here again there was a lot of duplicate code that was consolidated as this uses the `Mail` class to send and take care of PBE etc.

There is a basic abstract class to set linebreaks, transport and other things common to the above (3) new classes.  Everything is now in the Mail namespace.  I left wrapper functions in `Mail.subs.php` so old calls still work.  Did I break anything? most likely ;)

### DIFF
--- a/sources/ElkArte/Controller/ScheduledTasks.php
+++ b/sources/ElkArte/Controller/ScheduledTasks.php
@@ -14,6 +14,7 @@
 namespace ElkArte\Controller;
 
 use ElkArte\AbstractController;
+use ElkArte\Mail\QueueMail;
 
 /**
  * This controllers action handlers are automatically called.
@@ -87,14 +88,7 @@ class ScheduledTasks extends AbstractController
 	 */
 	public function action_reducemailqueue()
 	{
-		global $modSettings;
-
-		// Lets not waste our time or resources.
-		if (empty($modSettings['mail_queue_use_cron']))
-		{
-			// This does the hard work, it does.
-			require_once(SUBSDIR . '/Mail.subs.php');
-			reduceMailQueue();
-		}
+		// This does the hard work, it does.
+		(new QueueMail())->reduceMailQueue();
 	}
 }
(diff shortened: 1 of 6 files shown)


Spuds:
No comments, guess its OK to go !

PR #3611, opened 2022-03-08
Theme pick

Spuds:
This PR breaks out the user theme pick from the Admin controller (managethemes).

Previously when the user was choosing their theme of choice, from their profile, the code they were running was actually in the Admin conrtoller `ManageThemes.php` which contained various section they were allowed to run as long as they were not the Admin, yes that is how it was LOL  Anyway it was not ideal construct so now its a separate profile function and separate profile templates.   This does cause some amount of duplication but at least things are properly aligned.  

As always there were a few bugs stomped along the way, theme preview was not working, variant selection lists were wrong and always showed the default theme, etc


Spuds:
No comment, in it goes !

PR #3612, opened 2022-03-10
Break display controller into a series of methods

Spuds:
A start, and for now, and end to breaking up `Display.php` into a series of smaller methods.  

I did what I could to move all `$context` manipulation out of the main flow and instead directly use the `topicinfo` values.  The topicinfo values were being used  to set $context values and then use those $context values were used in various logic points, pretty messy.   Now once information are known, or computed, `$context` is loaded for the templates.  Still some rough spots as some subs don't provide return values but instead stuff $context with the result.

As always I'm sure I've caused some issue somewhere as the order of what is computed has changed to some degree, but I think its more readable now and in a place that real refinement can take place.

### DIFF
--- a/sources/ElkArte/Controller/Display.php
+++ b/sources/ElkArte/Controller/Display.php
@@ -34,22 +34,31 @@ use ElkArte\ValuesContainer;
 class Display extends AbstractController
 {
 	/** @var null|object The template layers object */
-	protected $_template_layers = null;
+	protected $_template_layers;
 
 	/** @var int The message id when in the form msg123 */
 	protected $_virtual_msg = 0;
-
-	/** @var null|\ElkArte\MessageTopicIcons Takes care of rendering the message icons */
-	protected $_icon_sources = null;
 
 	/** @var int Show signatures? */
 	protected $_show_signatures = 0;
 
 	/** @var int|string Start viewing the topics from ... (page, all, other) */
-	private $_start;
+	protected $_start;
+
+	/** @var bool if to include unapproved posts in the count  */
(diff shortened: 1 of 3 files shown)


emanuele45:
Random thought:
https://github.com/elkarte/Elkarte/pull/3612/files#diff-aed017bb41b71b3f6cd77616ea41ebfe859afe51a7f1c5e2e6585ddf3a81a550R278
wouldn't make sense to toggle these types of behaviours via a variable?
Something like:
```
theme()->addJavascriptVar(['enable_video' => !empty($modSettings['enableVideoEmbeding']) ? 'true' : 'false');
```
that way we don't have to assume much on the "source" side and we can load the videos the way we want on the javascript directly?
I guess there is a possible slightly overhead on the client-side, but....is it worth considering?

emanuele45:
I tried that path a couple of times but never managed to wrap it up, good job!!

Spuds:
Indeed, I also think that should be done.  There are a few other places in the code where that is done that way as well so should be done in a separate PR.  

I noticed that I boned up the ALL function a wee  bit so I'll make an update to this shortly.  As you know its hard to unravel some of these functions as they do so many crazy resets.

emanuele45:
Spuds wrote on 3/12/22 3:52 PM:
>
> Indeed, I also think that should be done. There are a few other places 
> in the code where that is done that way as well so should be done in a 
> separate PR.
>

Yep, yep indeed.
It was just that I noticed it there (again) and I realized that it made 
sense. :laughing:

Spuds:
Added the JS ideas as new issue.

PR #3613, opened 2022-03-12
Fixes, improvements and syntax sugar

Spuds:
Another one of these "cleanup" PR's that all hate but are needed from time to time to keep things as clean as possible.  @emanuele45 I know you love these !  Helps keeps our rebase skills sharp LOL

- removed a lot of redundant code in the validator
- a few theme related fixes/tweaks
- improve, maybe, the prefetch header detection
- lots of ?? sugar
- Some return optimization some as in ?? some as in don't set a value just to return it
- use of newer constants so we do not have to upper/lower/i search
- removal of ancient code
- docblock fixes and some missing {}'s
- enhanced the way custom profile fields were tracked when feature is turned on/off
- use of min/max functions to clean up crazy nested if/else stuff
- anything I noted along the way


emanuele45:
I'm starting to hate you. :joy:

Spuds:
I can understand that 😺

Spuds:
No conflicts so in it goes (so we can chase moar bugs!)

Issue #3614, opened 2022-03-13
Improve JS enablement logic/flow


Spuds:
Random thought:
https://github.com/elkarte/Elkarte/pull/3612/files#diff-aed017bb41b71b3f6cd77616ea41ebfe859afe51a7f1c5e2e6585ddf3a81a550R278
wouldn't make sense to toggle these types of behaviours via a variable?
Something like:
```
theme()->addJavascriptVar(['enable_video' => !empty($modSettings['enableVideoEmbeding']) ? 'true' : 'false');
```
that way we don't have to assume much on the "source" side and we can load the videos the way we want on the javascript directly?
I guess there is a possible slightly overhead on the client-side, but....is it worth considering?

_Originally posted by @emanuele45 in https://github.com/elkarte/Elkarte/issues/3612#issuecomment-1065869141_

emanuele45:
On that one, I did a little bit, but... yeah, facing the issue that it feels right to just move everything into some template call... and that may lead to a little bit of troubles. :laughing:

PR #3620, opened 2022-03-27
A bunch of fixes and partial addressing #3614

emanuele45:


Spuds:
Nice :D ... That helps clean up a bunch of areas.

Remember I blew up Display.controller so that is causing your merge conflicts.

emanuele45:
Oh, I thought I already rebased... oh well.

emanuele45:
I noticed you moved attachments to be exposed to triggers, there was a particular reason?

Spuds:
> I noticed you moved attachments to be exposed to triggers, there was a particular reason?

I was working on a new feature which will add some microdata to the templates, notably display.  Having access to the attachments allows to build a more rich set of micro data.  It specifically adds open graph and schema.org style data, one for when you share a post, and one to help improve search results.  The way I did it was to tie into the display events and so needed the attachments to be a class property so I could use them.  It only really needs access if there is an attachment on the first post as that is what is used to build the data.

emanuele45:
I forgot to mention I was working on your comment above... oh well, will open a new PR later!

Spuds:
Cool .. After I merged this, I was like, hey where can I get access to the attachment array 🙈 

The previous display render data had the inline attachments in those results and then the display class had the below post attachments in the the attachment array that I had placed as a property to gain access.  

Anyway glad you are working on it !

Spuds:
I had forgotten to enable the new attachment module, so now I see the attachment array in the display renderer.  The only thing that is missing really are the ILA attachments which are excluded from that array.  May just set a flag in that array, like `show_inline` and let the template exclude them from showing below the post vs not having the data passed at all?

emanuele45:
hmm... if that is the case, then, since I'm facing a little issue with my original idea, a simple solution could be to, instead of "skipping" the ILA attachments from the `$output['attachments']` array, to add an index like "ila" that would be then used at theme-level to ignore those that are presented in-line.
And probably, in a certain way, this is an even better solution, because it leaves the final decision of how to visualize the attachments to the theme (for example it could open the door to the possibility to present the ILA attachments in the download area as a simple link because reasons).

Spuds:
A separate array is good as it would keep old themes working as intended.

themes/default/Display.template.php
```diff
@@ -20,6 +20,10 @@
 function template_Display_init()
 {
 	theme()->getTemplates()->load('GenericMessages');
+	if (!empty($context['likes']))
```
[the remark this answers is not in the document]
emanuele45: IIRC Random is added at install-time, see:
```
				array('modules_display', 'mentions,poll,verification,random'),
				array('modules_post', 'mentions,poll,attachments,verification,random'),
```
and is not _meant_ to be disabled, so there should be no toggle anywhere.
It's technically an hard-coded module, always on. I just didn't think about actually hard-coding it and went with the _standard_ way of adding it to the database.

Random was intended as a module for small tasks that may not be worth having their own module (either for sake of mental health or jut as a temporary solution until a better place was found) but still nice to have out of the controllers.

sources/ElkArte/Modules/Random/Display.php
```diff
@@ -44,6 +45,7 @@ public static function hooks(EventManager $eventsManager)
 
 			add_integration_function('integrate_topic_query', '\\ElkArte\\Modules\\Random\\Display::followup_topic_query', '', false);
 			add_integration_function('integrate_display_message_list', '\\ElkArte\\Modules\\Random\\Display::followup_message_list', '', false);
+			add_integration_function('integrate_display_message_list', '\\ElkArte\\Modules\\Random\\Display::load_likes', '', false);
```
Spuds, line 48: Where that is located is contingent on `if (empty($modSettings['enableFollowup']))` which I think means followups have to be disabled for `load_likes` to be called.  I thing that should be moved to its onw likes_enabled check ?

sources/ElkArte/Modules/Random/Display.php
```diff
@@ -74,6 +76,23 @@ public static function followup_message_list($messages)
...
+	/**
+	 * Show likes.
+	 *
+	 * @param int[] $messages
+	 */
+	public static function load_likes($messages)
+	{
+		global $modSettings, $context, $txt;
+
+		if (!empty($modSettings['likes_enabled']) && !empty($messages))
```
Spuds, line 88: I we check for likes enabled in the section where load_likes is set as an integration function this could be simplified to 
`if (!empty($messages))`

sources/ElkArte/Modules/Random/Display.php
```diff
@@ -74,6 +76,23 @@ public static function followup_message_list($messages)
...
+		{
+			require_once(SUBSDIR . '/Likes.subs.php');
+			$context['likes'] = loadLikes($messages, true);
+			theme()->getLayers()->addBefore('load_likes_button', 'body');
```
Spuds, line 92: Which this being loaded, there should be no need to call the template_load_likes_button_above() from display template init ?

Spuds:
Ah, for some reason it was not that way on my install, I had to go and add it.  

So for the rest I put a couple comments in the code, but `template_load_likes_button()` as a function does not exist and `if (!empty($context['likes']))` is not set due to the way `random` is loading (or not) the integration function, which when called will properly load the likes template (_above) as defined in the GenericMessages template.  I hope that makes sense and its not that I'm going mental.

Spuds:
@emanuele45 Sorry to be 🐛 I'm looking at the ILA fix, where we return both an attachment data and the ila data array.  Thats easy enough but while working on that (so I can do a bit more on the micodata) I noticed a little issue.

The way the code is now set, it did not load attachments array if it was an ILA.  The problem is if its an unapproved attachment then you don't get the option to approve it from the display page  as the attachment is simply not listed. OK easy fix to not load/continue the ILA array unless its also approved, that way the attachment array still gets it, allowing easy approval.

No what should we show in the post for an unapproved ILA?  Right now I have it as our default mime image, or it could be as simple as an X mark, or maybe an empty outline box, or use a text to image or waiting approval?

Spuds:
Another thought is just the paperclip icon or simply nothing at all.

emanuele45:
Spuds wrote on 4/3/22 8:33 PM:
>
> @emanuele45 <https://github.com/emanuele45> Sorry to be 🐛 I'm looking 
> at the ILA fix, where we return both an attachment data and the ila 
> data array. Thats easy enough but while working on that (so I can do a 
> bit more on the micodata) I noticed a little issue.
>
> The way the code is now set, it did not load attachments array if it 
> was an ILA. The problem is if its an unapproved attachment then you 
> don't get the option to approve it from the display page as the 
> attachment is simply not listed. OK easy fix to not load/continue the 
> ILA array unless its also approved, that way the attachment array 
> still gets it, allowing easy approval.
>
> No what should we show in the post for an unapproved ILA? Right now I 
> have it as our default mime image, or it could be as simple as an X 
> mark, or maybe an empty outline box, or use a text to image or waiting 
> approval?
>
>
hmm... good point.
Another option could be a dedicated message:
```
         $img = new TextImage($text);
         $img = $img->generate(200);

         $this->prepare_headers('no_image', 'no_image', 'image/png', 
'inline', 'no_image.png', true, false);
         Headers::instance()->sendHeaders();
         echo $img;

         obExit(false);
```
Though it would imply some more server processing and it may be rather 
ugly. :D

An "x" or a clip may lead thinking there is a problem?
Maybe really a simple text message injected into the post saying 
something along the lines "when the image will be approved it will 
appear here"?

Spuds:
Nods ... I was also concerned about using a default image, or an x or clip as it would likely generate the ever useful "Hey your attachment is broken" responses.  The text2image is no worse then making a thumbnail   I'll see what may work and go from there!

emanuele45:
Spuds wrote on 4/3/22 1:10 AM:
> In sources/ElkArte/Modules/Random/Display.php 
> <https://github.com/elkarte/Elkarte/pull/3620#discussion_r841128990>:
> > @@ -74,6 +76,23 @@ public static function followup_message_list($messages)
>   		$context['follow_ups'] = followupTopics($messages, self::$includeUnapproved);
>   	}
>   
> +	/**
> +	 * Show likes.
> +	 *
> +	 * @param int[] $messages
> +	 */
> +	public static function load_likes($messages)
> +	{
> +		global $modSettings, $context, $txt;
> +
> +		if (!empty($modSettings['likes_enabled']) && !empty($messages))
> +		{
> +			require_once(SUBSDIR . '/Likes.subs.php');
> +			$context['likes'] = loadLikes($messages, true);
> +			theme()->getLayers()->addBefore('load_likes_button', 'body');
>
> Which this being loaded, there should be no need to call the 
> template_load_likes_button_above() from display template init ?
>
Indeed, this is what I was looking for, I didn't check that part of the 
reply before.
The part
```
     if (!empty($context['likes']))
     {
         template_load_likes_button();
     }
```

can be removed without issues.

emanuele45:
Spuds wrote on 4/3/22 1:09 AM:
> So for the rest I put a couple comments in the code, but 
> |template_load_likes_button()| as a function does not exist

Ohh... that's indeed another problem.
I think I see what happened here.
Let me try to fix it.
IIRC I changed my mind midway, fixed it somewhere else, but forgot tu 
push the fix (or something like this).

PR #3621, opened 2022-03-28
[WIP] Introduce recursive events: events can now be triggered from within modules

emanuele45:
@spuds there we go: recursive events... actually probably it would have been better to call them nested? Yeah. Whatever.
I thought thi was going to solve the issue with attachments you were facing, but it doesn't, though it looks pretty cool to me, so I just pushed it... let me know what you think. :laughing:

### DIFF
--- a/sources/ElkArte/EventManager.php
+++ b/sources/ElkArte/EventManager.php
@@ -42,8 +42,14 @@ class EventManager
 	/** @var object Instances of the controller. */
 	protected $_source = null;
 
+	/** @var object Instances of the modules. */
+	protected $_modules = [];
+
 	/** @var string[] List of classes already registered. */
 	protected $_classes = array();
+
+	/** @var int Remembers the depth of nested modules triggers. */
+	protected $depth = 0;
 
 	/** @var null|string[] List of classes declared, kept here just to
 	    avoid call get_declared_classes at each trigger */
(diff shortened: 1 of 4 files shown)


Spuds:
I think this is very cool .... When I was first playing with the micro data I was thinking it would be nice if I could trigger an event from the code !

emanuele45:
Spuds wrote on 3/29/22 12:31 AM:
>
> I think this is very cool .... When I was first playing with the micro 
> data I was thinking it would be nice if I could trigger an event from 
> the code !
>

Little question: while working on this, I assumed that the trigger in 
the addon, should pass the states of the addon itself, not those of the 
controller, reason is the "addon of the addon" should be extending the 
addon and not the controller, otherwise it should hook up into the 
controller itself.
@spuds do you think the reasoning works or should it always pass the 
states and variables of the controller?

Spuds:
I *think* that makes sense, I'd be able to give a better answer if (when) I work on an addon that utilizes the functionality, but rolling it around in my head extending the addon seems proper.

tenbuddy:
Thank you for your kind information. Happy to learn from you all. It'd be cool if you get the [latest news](https://www.oracleglobe.com/) update.

Spuds:
Don't think this is going anywhere now !

PR #3622, opened 2022-04-06
Updates to ILA

Spuds:
This started as a simple return both a below post attachment array and the ILA array.  

Next was the ILA code was not handling non approved attachments in a graceful way (simply ignored them, left bad attach code in place), this now will present a pending approval image.  Renamed the `action_no_attach` method to `action_text_to_image`, since its used for both attachment missing and awaiting activation string, and added some features.

Next while testing ILA, there were a few edge problems, like if someone used a width attribute on a file attachment, plain `[attach]tags[/attach]` were not working.  detecting the no approved status was problematic, and a several other tag creation issues.  

To fix the problems and not make the ILA code overly complex, and still not working in all cases, I chose to add a new feature to the `BBC Parser`. to overcome the way the parser builds the tags.  Short explanation:

For a tag with parameters say `[attach width=300]x[/attach]`,  First the parameter attr values are determined, like `width=300` and then it will do a swap/replace to replace any `{width}` tags in the defined `ATTR_CONTENT`.  Next, it will run any `ATTR_VALIDATE` function and take those results and substitute for any `$1` strings in the `ATTR_CONTENT` value.   This allows a context tag like `...` to render.  

The _short coming_ is for something like ILA, which has to deal with images and files and thumbnails and links, a single `ATTR_CONTENT` can not be built.  Currently the ILA the tag is a confusing template of `$1's and {parms}` which had sneaky open quotes but when all was combined made a valid html tag.  But this has major short comings when you have to build image and non image tags.  You could make `ATTR_CONTENT`  just be $1 and build the entire tag in the validation function, but the short coming there is you can't get at the `{param}` values the parser created since they don't exist as substituion strings in `ATTR_CONTENT` .

To help with the above I added a new code attribute to the Code class that Parsebbc will use.  The new Code param is `Codes::ATTR_RESET]` which will take a string like `{width}~{height}~{type}`  This allows the true final context tag to be dynamically set. 

What happens internally, Parsebbc will render this "reset" tag as the tags `ATTR_CONTENT` and then pass the results to any `ATTR_VALIDATE` function (defined by the bbc tag) which in turn can use them to build a final `ATTR_CONTENT` which simply points back to the reference &variable we set.  I think you could just use `$1` as well and skip having the reference pointer and just return the completed tag ( I may check that now).  

Now each `[attach]` tag gets a `Codes::ATTR_RESET]` value that corresponds to the parameters it accepts, parsebbc does its thing and parses those values then sets them to the tag being rendered.  That rendered tag is passed to the validation function with itself returns the completed string back to `ATTR_CONTENT`  which then becomes the rendered tags content back in parsebbc.

**TLDR** I added a single line of code to parsebbc and a new optional attr tag to Code:: which allows us to remove a bunch of crud in ILA and make it more robust.  I wanted to detail the why it was done since I will forget by next week!

### DIFF
--- a/sources/ElkArte/AttachmentsDisplay.php
+++ b/sources/ElkArte/AttachmentsDisplay.php
@@ -15,7 +15,7 @@ namespace ElkArte;
 
 class AttachmentsDisplay
 {
-	/** @var The good old attachments array */
+	/** @var array The good old attachments array */
 	protected $messages = [];
 
 	/** @var mixed[] The good old attachments array */
@@ -29,7 +29,7 @@ class AttachmentsDisplay
 	 * @param int[] $posters
 	 * @param bool $includeUnapproved
 	 */
-	public function __construct($messages, $posters,  $includeUnapproved)
+	public function __construct($messages, $posters, $includeUnapproved)
 	{
 		$this->messages = $messages;
 		$this->includeUnapproved = $includeUnapproved;
@@ -63,7 +63,6 @@ class AttachmentsDisplay
 	 * @param string|null $filter name of a callback function
 	 * @param mixed[] $all_posters
 	 *
-	 * @return array
 	 * @package Attachments
 	 */
 	protected function getAttachments($messages, $includeUnapproved = false, $filter = null, $all_posters = array())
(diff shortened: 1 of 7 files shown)


Spuds:
Saving you the 😧 When a better / cleaner way is thought of we can make a new PR

emanuele45:
A bit late to the party, but :+1: !

PR #3629, opened 2022-05-16
Mostly Theme update / tweaks

Spuds:
Another batch of random

- pulls in the chart js stuff for stats etc. to replace the old bar graphs
- replace stats donut hole PNG file with on the fly SVG icons
- lots more CSS cleanup with some added ally fixes and some rtl fixes
- dropping of google js minifity as the service has been depreciated
- update of minify to allow on individual files, allowing combined/non, combined/min, individual/non, individual/min
- output js defer in head section as browser support is ubiquitous
- update quotes to be blockquote with cites with test and more CSS updates to support
- change allow/not nested quotes to be a level e.g. 2 deep vs only on/off
- allow side by side video embed and add more (experimental) sites to support.
- Bugs and stuff


Spuds:
Comments ... leave them on the site LOL

PR #3638, opened 2022-06-17
Add basic Markdown support to the message parser

Spuds:
We can call this experimental as who know what edge cases will show up.  This allows us to use some basic markdown syntax in the editor and have it render.  Added are:

`**Text** or __Text__ => [b]Text[/b]`
`*Text* or _Text_ => [i]Text[/i]`
`~~Text~~ => [s]Text[/s]`
` 'Text' => [icode]Text[/icode]`
` '''Text''' => [code]Text[/code]`
`> Text => [quote]Text[/quote]`
`--- or *** or ___ => [hr]`

The code ones are really backticks `, could not escape it so used single quote ' in the above list.

The major issue with code, inline or fenced, is that it will not be cleaned by preparsebbc, so if its wrong, its wrong.  🤷‍♂️ I suppose it could be added.


Spuds:
Nothing? ... welp guess we can give it a whirl 😞

PR #3649, opened 2022-07-11
delete empty paragraph

interlab:


Spuds:
Yup, can't nest <p> inside of a <p> !  Plus that is already wrapped in a div ... Good catch, thanks!

PR #3660, opened 2022-07-28
Check if shell_exec is disabled on functions list

Trekkie101:
Check if shell_exec is disabled before attempting to use it to avoid the PHP 8.1 error.


Spuds:
Thanks for the report and fix !

PR #3663, opened 2022-08-09
Another PR of fixes

Spuds:
Too many one off fixes to write a lengthy description that no one reads anyway.  This fixes various issues reported on the site or in our issues tracker.  Also added a few items as detailed in the issue tracker requests etc.  See the individual commits for the details.


Spuds:
Nice work, looks great, merge it yourself and get a reward !

PR #3668, opened 2022-09-10
Smiley and Message Icon updates

Spuds:
This makes several changes to the smileys and message icon functions

- Allows the use of SVG for smileys
- Replaces default set with SVG and expands the supplied collection.
- Default set will follow the chosen emoji library, if the library is changed, the smiley will follow 
- Moves / integrates emoji processing along side the smiley vs pre/post bbc
- Allows setting custom emoji tags for site specific image
- Allows overwriting a standard emoji with one of the sites choosing
- Smiley sets can use different file types across sets
- Can define the use of a library / site emoji in a smiley set with no file and have it use the library svg image
- Updated message icon images so custom and default match.
- Centralized the standard message icon array, it was in a few spots
- Updated the installer and upgrader.  Upgrade, TBH, is a bit of a crap shoot, does the best it can
- Removed option for user to select their own smiley set and added the option to simply turn them off (per theme user setting)

As always various cleanup of files and more css fixes/adjustments. updated to latest jquery.  Also removed the option to choose your version as the library is so stable now.


Spuds:
Yeah! 😓on this one.

PR #3708, opened 2023-04-10
First commit of redis cache utilising predis

tinoest:
@Spuds not sure if you want this as part of the core, or if I am doing this the right way. 

The ip for the redis server is hardcoded currently, so if you do a checkout edit this line to what your ip is to get it working. 

`$this->obj	= new \Predis\Client('tcp://172.19.0.5:6379');`

I will change and work out how to add multiple different servers if you are happy using predis. It _seems_ to work ok otherwise.


Spuds:
I picked the `Redis.php` file from this PR and renamed it to `Predis.php` because I also added a method to use the pecl `phpredis` extension.  

I'm still unsure what to do about the predis libray itself.  I'm inclined to add that to `composer` and start using the vendor library in place of the `ext` library going forward.  Then this would be added to the suggest section and the core would only have the single interface file to translate elk speak to predis speak.

Also sorry this took me so long to get to, life happens!

Spuds:
For completeness, the new PR is https://github.com/elkarte/Elkarte/pull/3801

It has both methods and also adds (back again) the cache password and UID values as `Redis` can have `auth()` enabled so we need to support that.  I did not add that to the predis method just yet but did test it with phpredis

Spuds:
Closed as this was incorporated in https://github.com/elkarte/Elkarte/pull/3801

PR #3819, opened 2025-12-03
Some install (and update) improvements

Spuds:


Spuds:
Created new one as this had some postgre issue

